PR #6, opened 2018-12-06
DASD-4445 Cosmos DB with Gremlin API

rivoric:
Added another Cosmos DB this time with Gremlin API to Azure resources


rivoric:
Master was not at latest

PR #39, opened 2018-12-19
fix tests

sbradley090877:


src/Dfc.CourseDirectory.Services/PostCodeSearchCriteria.cs
```diff
@@ -8,38 +8,18 @@ namespace Dfc.CourseDirectory.Services
...
 
         protected override IEnumerable<object> GetEqualityComponents()
         {
-            yield return Search;
-            //yield return Top;
-            //yield return Skip;
-            //yield return Count;
-            //yield return Filter;
-            //yield return Facets;
+            yield return Search;;
```
LiamWright: Double semi colon

MahmoodAkhtar: @sbradley090877 I think you might need to pull from master again because Anatoli's PR has just gone in before yours.

PR #90, opened 2019-01-04
Dfc 6323 your courses line 2 view component

sbradley090877:


src/Dfc.CourseDirectory.Web/Controllers/CoursesController.cs
```diff
@@ -49,141 +49,82 @@ public class CoursesController : Controller
...
+                    CreatedDate = Convert.ToDateTime("2019-01-03T11:17:02.514746+00:00"),
+                    CreatedBy = "ProviderPortal-AddCourse",
+                    UpdatedDate = Convert.ToDateTime("0001-01-01T00:00:00"),
+                    UpdatedBy = null
+                }
             };
+           
+            
             Course[] course = new Course[]
             {
```
LiamWright: Can you please copy this stub data over so the page looks more populated until we get the API sorted?

PR #106, opened 2019-01-09
DFC-6320 - Your courses service wrapper

ian-findacourse:


src/Dfc.CourseDirectory.Web/Controllers/CoursesController.cs
```diff
@@ -473,7 +344,7 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
         {
             var selectVenue = new SelectVenueModel
             {
-                LabelText = "Select course venue",
+                LabelText = "Select course venue"
```
LiamWright: Does this need a comma?

PR #171, opened 2019-02-04
add hours to dropdown

sbradley090877:


src/Dfc.CourseDirectory.Web/ViewComponents/Courses/Duration/Duration.cs
```diff
@@ -3,13 +3,28 @@
...
+            foreach (DurationUnit eVal in DurationUnit.GetValues(typeof(DurationUnit)))
+            {
+                if (eVal.ToString().ToUpper() != "UNDEFINED")
+                {
+                    var item = new SelectListItem { Text = WebHelper.GetEnumDescription(eVal) };
+                    model.DurationUnits.Add(item);
+                }
+            };
+
+            model.DurationUnits.Insert(0,model.DurationUnits.FirstOrDefault(x => x.Text.ToUpper()=="HOURS"));
```
MakCom: If we had hours as 1, we would not need to do these lines of code.
And we have to repeat it every time we have DurationUnit dropdown.

PR #175, opened 2019-02-06
NCSD-126 Updated to new templates

rivoric:
Changed all nested templates to dfc-devops ones


Resources/template.json
```diff
@@ -154,24 +154,30 @@
...
+      "metadata": {
+        "description": "Resource group for the key vault - only used if appHostName is specified"
       }
     },
     "certificateName": {
       "type": "string",
       "defaultValue": "",
       "metadata": {
-        "description": "Certificate for the custom host name - only used if appHostName is specified"
+          "description": "Array of certificates - only used if appHostName is specified"
```
NickGraham101: Should this have been changed?  Looks like it's still the name of the certificate for the custom hostname

PR #247, opened 2019-03-01
Dfc 7295 view component for levels

MahmoodAkhtar:


src/Dfc.CourseDirectory.Web/Controllers/YourCoursesController.cs
```diff
@@ -0,0 +1,210 @@
...
+        {
+            if (courses == null) return string.Empty;
+            if (string.IsNullOrWhiteSpace(id)) return string.Empty;
+
+            var found = courses.FirstOrDefault(x => x.Id == id)?.QualificationTitle;
+
+            return found;
+        }
+
+        public async Task<IActionResult> Index(
```
suniara: do we need an Authorise attribute here?

PR #791, opened 2019-07-22
Checking for values in addition to status

LiamWright:


LiamWright:
Includes changes from COUR-1413

PR #1120, opened 2019-12-04
BulkUpload message for big files

MTSarwar:


src/Dfc.CourseDirectory.Web/ViewComponents/Dashboard/Dashboard.cs
```diff
@@ -94,7 +94,13 @@ public async Task<IViewComponentResult> InvokeAsync(DashboardModel model)
 
                 IEnumerable<Course> inValidCourses = courses.Where(c => c.IsValid == false);
 
-              
+                var getApprenticeshipResult = _apprenticeshipService.GetApprenticeshipByUKPRN(UKPRN.ToString()).Result;          
+
+               
+
+                var ApprenticeshipBulkUploadReadyToGoLive = _apprenticeshipService.GetApprenticeshipByUKPRN(UKPRN.ToString()).Result.Value.Where(x => x.RecordStatus == RecordStatus.BulkUploadReadyToGoLive);
```
gunndabad: `await` here instead of `.Result`

MTSarwar: it required bigger change

PR #1137, opened 2020-01-29
validation for missing Standard Code and standard version

MTSarwar:


src/Dfc.CourseDirectory.Services/BulkUploadService/ApprenticeshipBulkUploadService.cs
```diff
@@ -174,12 +174,20 @@ private IStandardsAndFrameworks Mandatory_Checks_GetStandard(IReaderRow row)
...
+                    throw new BadDataException(row.Context, $"Validation error on row {row.Context.Row}. Missing Standard Code.");
+                }
 
-                if (!standardCode.HasValue && standardVersion.HasValue)
+                else if (!standardCode.HasValue && standardVersion.HasValue)
                 {
                     throw new BadDataException(row.Context, $"Validation error on row {row.Context.Row}. Missing Standard Code.");
                 }
-                
+                else if (!standardCode.HasValue && !standardCode.HasValue)
```
gunndabad: Duplicate check here..?

MTSarwar: this case both fields are empty.

MTSarwar: Excellet Catch !!

PR #1325, opened 2020-03-28
Store data protection keys in Blob Storage

gunndabad:


src/Dfc.CourseDirectory.Web/Startup.cs
```diff
@@ -227,18 +224,43 @@ public void ConfigureServices(IServiceCollection services)
...
+                var cloudStorageAccount = new CloudStorageAccount(
+                    new Microsoft.Azure.Storage.Auth.StorageCredentials(
+                        Configuration["BlobStorageSettings:AccountName"],
+                        Configuration["BlobStorageSettings:AccountKey"]),
+                    useHttps: true);
+
+                var blobClient = cloudStorageAccount.CreateCloudBlobClient();
+
+                var container = blobClient.GetContainerReference(Configuration["DataProtection:ContainerName"]);
+                container.CreateIfNotExists();
```
gunndabad: @thelious1974 better to do this in ARM template?

thelious1974: @gunndabad Yes, it is..   it's defined in the shared resources arm templates.  Let me know what you want it called and I'll add a task to get it added in today

PR #1433, opened 2020-04-27
added missing handler.

MrKevJoy:
- Added missing handler.


src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/UpdateApprenticeshipHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+                    ProviderUKPRN = request.ProviderUkprn,
+                    Radius = l.Radius,
+                    RecordStatus = 1,
+                    Regions = l.Regions,
+                    UpdatedBy = request.CreatedByUser.Email,
+                    UpdatedDate = request.CreatedDate,
+                    VenueId = l.VenueId
+                }).ToList(),
+                RecordStatus = 1,
+                CreatedDate = request.CreatedDate,
```
gunndabad, line 54: These two `Created` fields shouldn't be overwritten (they shouldn't really be part of command at all)

MrKevJoy: removed these now, and updated the test handler too.

PR #1468, opened 2020-05-12
Add an attribute to indicate slow tests so they easily be skipped

gunndabad:


soharadesign:
We should build some sort of common lib nuget package for all of these - I could use this myself!

gunndabad:
Hoping once we consolidate repos down a bit it will be a mere project reference away :-)

PR #1478, opened 2020-05-14
Add visually hidden text to QA pages for accessibility

gunndabad:


gunndabad:
This is what the GDS site says to do

PR #1500, opened 2020-06-01
Fix total number of pages in LARS search results (PTCD-296)

MTSarwar:
total number is equal to total OData divided by items per page


src/Dfc.CourseDirectory.Web/Controllers/UnregulatedCoursesController.cs
```diff
@@ -268,7 +268,7 @@ public async Task<IActionResult> ZCodeNotKnown([FromQuery] ZCodeNotKnownRequestM
                 model.Url = Request.GetDisplayUrl();
                 model.PageParamName = _larsSearchSettings.PageParamName;
                 model.ItemsPerPage = _larsSearchSettings.ItemsPerPage;
-                model.TotalCount = result.Value.ODataCount ?? 0;
+                model.TotalCount = result.Value.ODataCount / model.ItemsPerPage ?? 0;
```
gunndabad, line 271: This isn't correct - `TotalCount` should be total number of items not total number of pages (calculation for that is incorrect at any rate)

MTSarwar: now calculating before invoking VC

PR #1534, opened 2020-06-19
Enable diagnostic messages for Core.Tests

gunndabad:


timabell:
I wonder if it would be better to duplicate the file rather than referencing up a folder

gunndabad:
> I wonder if it would be better to duplicate the file rather than referencing up a folder

My thinking is that we want consistent behaviour across both test projects currently

PR #1547, opened 2020-06-26
PTCD-414 - Updated design and text for Add Course page

timabell:
* Strip trailing whitespace.
* Using back-link tag helper from pttcd instead of nuget after talking to James.
* The page H1 in the jira ticket overrules the one in the docx attached to the ticket.
* All copy changes are from the Jira ticket.

Note the previous page loses state when going back to it with the back link but that's beyond the scope of this ticket.


timabell:
rebased on `dev`

PR #1555, opened 2020-07-02
PTCD-410 - Refactor - Move back link to toggleable layout option

timabell:
* DRY markup for back links by moving it into the layout and making it togglable
* Introduce a layout for strong typing
* Introduce extension method for clean syntax for storing layout model
* Introduce new test project for Web(v1) to use as we work on the existing code

Used IDictionary rather than ViewDataDictionary for testability.

### DIFF
--- a/tests/Dfc.CourseDirectory.Web.Tests/LayoutModelStorageTests.cs
+++ b/tests/Dfc.CourseDirectory.Web.Tests/LayoutModelStorageTests.cs
@@ -1,0 +1,48 @@
+using System;
+using System.Collections.Generic;
+using Dfc.CourseDirectory.Web.Views.Shared;
+using Xunit;
+
+namespace Dfc.CourseDirectory.Web.Tests
+{
+    public class LayoutModelStorageTests
+    {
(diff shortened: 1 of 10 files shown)


tests/Dfc.CourseDirectory.Web.Tests/LayoutModelStorageTests.cs
```diff
@@ -0,0 +1,49 @@
+using System;
+using System.Collections.Generic;
+using Dfc.CourseDirectory.Web.Views.Shared;
+using FluentAssertions;
```
gunndabad, line 4: We don't use FluentAssertions anywhere else currently AFAIK. For me we should use xUnit straight until we've committed to FluentAssertions everywhere.

timabell: stripped out and forcepushed :+1:

timabell:
FluentAssertions removed from patch to reduce novelty and extra dependencies while we have so much else in flux.

PR #1560, opened 2020-07-06
PTCD-418 - Add back link to add-course summary

timabell:
screenshot showing back link and developer tools showing PRG requests


timabell:
rebased

PR #1566, opened 2020-07-07
PTCD-414 - Keep page state on navigate back to qualification

timabell:
The `no-store` `cache-control` header was causing the browser to lose the state of the qualification selection pages when navigating back from the following page.

I propose that because Course directory doesn't deal in any sensitive data that we don't need to worry about generated pages being stored in the user's browser cache, which is already reasonably well protected.

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Startup.cs
+++ b/src/Dfc.CourseDirectory.Web/Startup.cs
@@ -287,7 +287,6 @@ namespace Dfc.CourseDirectory.Web
                   new Microsoft.Net.Http.Headers.CacheControlHeaderValue()
                   {
                       NoCache = true,
-                      NoStore = true,
                       MustRevalidate = true,
                   };
                 context.Response.Headers[Microsoft.Net.Http.Headers.HeaderNames.Vary] =


gunndabad:
Given that we don't know why this was added, I'm nervous about taking it away.

timabell:
I am also nervous, but I think it is right here.

It was added in #77 as a fairly broad set of security modifications with no explanation of why this particular thing was needed.

The research I did turned up the following:

> “The no-store directive applies to the entire message and indicates that the cache must not store any part of the response or any request that asked for it.”
~ https://owasp.org/www-community/OWASP_Application_Security_FAQ

> “checking that for every page that contains sensitive information the server instructed the browser not to cache any data”
~ https://owasp.org/www-project-web-security-testing-guide/latest/4-Web_Application_Security_Testing/04-Authentication_Testing/06-Testing_for_Browser_Cache_Weaknesses

timabell:
I should also note that this is only one of the cache control headers, there are two others so this is not completely re-enabling caching. I think this just controls whether the browser can store a copy of the retrieved resource(s).

gunndabad:
> I should also note that this is only one of the cache control headers, there are two others so this is not completely re-enabling caching. I think this just controls whether the browser can store a copy of the retrieved resource(s).

There are some flows on the site (some of the QA-related journeys) where showing old cached content is not desired. We would be going backwards if for those things if we merge this change.

timabell:
Closing on the basis of @gunndabad 's comment

> There are some flows on the site (some of the QA-related journeys) where showing old cached content is not desired. We would be going backwards if for those things if we merge this change.

PR #1574, opened 2020-07-14
Replace the generic invalid ModelState filter with a more specific one

gunndabad:
We don't always want to 'early out' when there's a ModelState error.


src/Dfc.CourseDirectory.WebV2/MultiPageTransaction/InvalidMptxInstanceContextActionFilter.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class InvalidMptxInstanceContextActionFilter : IActionFilter
+    {
+        private static readonly Type _instanceContextType = typeof(MptxInstanceContext<>);
+        private static readonly Type _instanceContextWithParentType = typeof(MptxInstanceContext<,>);
+
+        public void OnActionExecuted(ActionExecutedContext context)
+        {
+        }
+
+        public void OnActionExecuting(ActionExecutingContext context)
```
timabell, line 17: Suggestion: the contents of this method is a little hard for me to parse, particularly being less familiar with action filters; it might be good to break out a couple of tiny well named methods to make it clearer what it's doing. The use of comment might also be a hint that this is needed. I think that the comment could possibly be completely replaced by good method call names which would be a small win.

src/Dfc.CourseDirectory.WebV2/MultiPageTransaction/InvalidMptxInstanceContextActionFilter.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+
+namespace Dfc.CourseDirectory.WebV2.MultiPageTransaction
+{
+    public class InvalidMptxInstanceContextActionFilter : IActionFilter
```
timabell, line 8: Shouldn't this new class have matching unit tests?

gunndabad: It’s covered by other higher level tests

PR #1583, opened 2020-07-16
Update Content + Layout on Apprenticeship Bulk Upload Page (PTCD 539)

jameslawesynergy:
image

Updated BulkUploadApprenticeships view to include new layout and copy.

Design updated:

image

Further designs updates (22/07/2020):
image

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Views/BulkUploadApprenticeships/Index.cshtml
+++ b/src/Dfc.CourseDirectory.Web/Views/BulkUploadApprenticeships/Index.cshtml
@@ -69,7 +69,12 @@
 
         <form method="post" id="bulkUploadForm" enctype="multipart/form-data" asp-controller="BulkUploadApprenticeships" asp-action="Index">
 
-            <p class="govuk-body">You can upload a CSV file to update your apprenticeship training courses, or you can @Html.ActionLink("add", "Index", "Apprenticeships"), @Html.ActionLink("edit or delete", "Index", "ProviderApprenticeships") each course individually.</p>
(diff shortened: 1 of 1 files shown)


soharadesign:
The line limit is there as we start seeing Cosmos issues on files around 400 lines - 375 is a little bit under that to give a bit of headroom. This number is dependent on the RU limit on Cosmos as far as I'm aware - once we get some optimisations in place this should disappear.

src/Dfc.CourseDirectory.Web/Views/BulkUploadApprenticeships/Index.cshtml
```diff
@@ -69,7 +69,12 @@
 
         <form method="post" id="bulkUploadForm" enctype="multipart/form-data" asp-controller="BulkUploadApprenticeships" asp-action="Index">
 
-            <p class="govuk-body">You can upload a CSV file to update your apprenticeship training courses, or you can @Html.ActionLink("add", "Index", "Apprenticeships"), @Html.ActionLink("edit or delete", "Index", "ProviderApprenticeships") each course individually.</p>
+            <p class="govuk-body">You can upload a CSV file to update your apprenticeship training courses, or you can:</p>
+
+            <ul class="govuk-list govuk-list--bullet">
+                <li>@Html.ActionLink("add a new", "Index", "Apprenticeships") apprenticeship training course</li>
```
timabell, line 75: That link text doesn't make sense out of context, which is how a screen reader user would hear it. I understand screen reader users tab through all the links, and the reader reads out all the link texts, without any of the surrounding text.

PR #1590, opened 2020-07-21
PTCD-508 - Add test coverage for apprenticeship bulk upload service

timabell:
These are "characterization tests" to show the current behaviour of the
system without introducing modifications, with a view to being able to
confidently change and refactor the code later.

* Add parameterless constructor to data objects to simplify testing.
* Make venue name settable so we can use an object initializer.

Written by Tim, with some pairing from James & James.

This is not complete coverage, there is plenty more to do (some todos in the PR), but it's a good start, and I think worth getting in as-is as a base for extending as desired.

Areas that need more work:

* testing exactly what is passed on to the next link in the chain
* what does the background upload process do, does it share the code?
* testing of any transformations made on the way through

Selection_20200721-02-0

### DIFF
--- a/Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
+++ b/Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
@@ -64,7 +64,7 @@ namespace Dfc.CourseDirectory.Services.Tests.BulkUploadService.Apprenticeship
             return this;
         }
 
-        public Stream Build()
+        public Stream BuildStream()
         {
             return new MemoryStream(Encoding.UTF8.GetBytes(BuildString()));
         }
--- a/Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
+++ b/Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
@@ -1,0 +1,101 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Dfc.CourseDirectory.Services.Tests.BulkUploadService.Apprenticeship
+{
+    /// <summary>
+    /// Builder Pattern implementation with fluent api for building arbitrary variations of csv data for testing bulk upload.
+    /// </summary>
+    internal class ApprenticeshipCsvBuilder
+    {
+        private bool _emptyFile;
+        private bool _generateInvalidHeader;
+        private bool _trailingNewline = true;
+        private string _newLine = "\r\n";
+        private readonly List<string> _rows = new List<string>();
+
+        private string CsvHeader()
+        {
+            var csvHeader = string.Join(",", ApprenticeshipCsvStructure.Fields);
+            if (_generateInvalidHeader)
+            {
+                return "modified_first_header_name_" + csvHeader;
+            }
+            return csvHeader;
+        }
+
+        /// <summary>
+        /// Lambda based API for adding rows.
+        /// To add a default valid row:
+        ///   `.WithRow()`
+        /// To add a customised row:
+        ///   `.WithRow(row => row.WithStandardCode("this-is-not-an-integer"))`
+        /// </summary>
+        public ApprenticeshipCsvBuilder WithRow(Action<ApprenticeshipCsvRowBuilder> configureRow = null)
+        {
+            var csvRowBuilder = new ApprenticeshipCsvRowBuilder();
+            configureRow?.Invoke(csvRowBuilder);
+            _rows.Add(csvRowBuilder.Build());
+            return this;
+        }
+
+        public ApprenticeshipCsvBuilder WithoutRows()
+        {
+            // This method isn't strictly necessary, you can just add no rows,
+            // but it makes the intent of the caller more explicit.
+
+            _rows.Clear();
+            return this;
+        }
+
+        public ApprenticeshipCsvBuilder WithUnixLineEndings()
+        {
+            _newLine = "\n";
+            return this;
+        }
+
+        public ApprenticeshipCsvBuilder WithoutTrailingNewline()
+        {
+            _trailingNewline = false;
+            return this;
+        }
+
+        public ApprenticeshipCsvBuilder WithInvalidHeader()
+        {
+            _generateInvalidHeader = true;
+            return this;
+        }
+
+        public ApprenticeshipCsvBuilder EmptyFile()
+        {
+            _emptyFile = true;
+            return this;
+        }
+
+        public Stream Build()
+        {
+            return new MemoryStream(Encoding.UTF8.GetBytes(BuildString()));
+        }
+
+        private string BuildString()
+        {
+            if (_emptyFile)
+            {
+                return "";
+            }
+
+            var sb = new StringBuilder();
+            sb.Append(CsvHeader());
+            sb.Append(_newLine);
+            sb.Append(string.Join(_newLine, _rows));
+            if (_trailingNewline)
+            {
+                sb.Append(_newLine);
+            }
+
+            return sb.ToString();
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    internal class ApprenticeshipCsvBuilder
+    {
+        private bool _emptyFile;
+        private bool _generateInvalidHeader;
+        private bool _trailingNewline = true;
+        private string _newLine = "\r\n";
+        private readonly List<string> _rows = new List<string>();
+
+        private string CsvHeader()
```
gunndabad, line 19:
```suggestion
        private string CreateCsvHeader()
```

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
```diff
@@ -0,0 +1,101 @@
...
+            return this;
+        }
+
+        public ApprenticeshipCsvBuilder EmptyFile()
+        {
+            _emptyFile = true;
+            return this;
+        }
+
+        public Stream Build()
```
gunndabad, line 77:
```suggestion
        public Stream BuildStream()
```

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
```diff
@@ -0,0 +1,63 @@
...
+        public ApprenticeshipCsvRowBuilder()
+        {
+            SetValidDefaultValues();
+        }
+
+        public ApprenticeshipCsvRowBuilder With(string field, string value)
+        {
+            if (!_values.ContainsKey(field))
+            {
+                throw new ArgumentOutOfRangeException($"Unknown field '{field}'");
```
gunndabad, line 27:
```suggestion
                throw new ArgumentOutOfRangeException(field, $"Unknown field '{field}'.");
```

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            _values[field] = value;
+            return this;
+        }
+
+        public ApprenticeshipCsvRowBuilder WithStandardCode()
```
gunndabad, line 34: Why isn't this parameterised by standard code + version?

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            return With("STANDARD_CODE", "122").With("STANDARD_VERSION", "1");
+        }
+
+        public ApprenticeshipCsvRowBuilder WithFrameworkCode()
```
gunndabad, line 39: As above

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvStructure.cs
```diff
@@ -0,0 +1,29 @@
+using System.Collections.Generic;
+
+namespace Dfc.CourseDirectory.Services.Tests.BulkUploadService.Apprenticeship
+{
+    internal static class ApprenticeshipCsvStructure
+    {
+        public static readonly List<string> Fields = new List<string>
```
gunndabad, line 7:
```suggestion
        public static readonly IReadOnlyList<string> Fields = new List<string>
```

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <summary>
+    /// Builder Pattern implementation with fluent api for building arbitrary csv row data for bulk upload testing
+    /// </summary>
+    internal class ApprenticeshipCsvRowBuilder
+    {
+        public const string DefaultTestVenueName = "A Test Venue";
+
+        private readonly Dictionary<string, string> _values = new Dictionary<string, string>(
+            ApprenticeshipCsvStructure.Fields.Select(
+                fieldName => new KeyValuePair<string, string>(fieldName, "")));
```
gunndabad, line 16:
```suggestion
        private readonly Dictionary<string, string> _values = ApprenticeshipCsvStructure.Fields.ToDictionary(f => f, _ => string.Empty);
```

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
gunndabad: Why do we have With* methods for these two fields only?

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
timabell: Done, thanks

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvBuilder.cs
timabell: Done, thanks, that is indeed clearer

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvStructure.cs
timabell: Ah, fancy, I haven't looked at the new types much. Thanks! Done

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
timabell: Nice, thanks. Done

timabell: I kept `""` because personally I find it easier to mentally parse

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/ApprenticeshipCsvRowBuilder.cs
timabell: > Why isn't this parameterised by standard code + version?

`WithStandardCode()` and `WithFrameworkCode()` are convenience methods for getting a basic valid record of either type. It keeps most of the tests short while avoiding the "mystery guest" of default values. The tests that test variations of codes, versions etc (e.g. `TestValidateAndUploadCSV_WithoutStandardCode_WithStandardVersion_Throws`) don't use these and instead directly set all the fields in various combinations which is pretty clear in my opinion. I'm pretty happy with the way the resultant tests read at the moment.

> Why do we have With* methods for these two fields only?

I started adding `With_____()` for each field, but it started to look like it was going to be a lot of extra code for little extra benefit compared to just passing the field name. I think we'll add more `With` methods if we need more non-trivial setup, but for what I've got covered so far nothing more was needed.

I should add that I've used YAGNI and DRY principles to keep it light as I can, but that equally this can and should be refactored and modified as the needs of the tests evolve. I haven't tried to predict what might be useful in future.

PR #1595, opened 2020-07-27
[PTCD-508] Add bulk-upload test coverage relevant to upcoming change

timabell:
This adds tests showing how multiple csv rows are combined into individual apprenticeship records with multiple child locations.

It also adds more detailed coverage around delivery modes, which is where a change in behaviour has been requested.

* Move theory attributes to be next to the test method signature to make it clearer that it is describing the method as defining a theory.


Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/LocationVariationFact.cs
```diff
@@ -0,0 +1,24 @@
+using Dfc.CourseDirectory.Core.Models;
+
+namespace Dfc.CourseDirectory.Services.Tests.BulkUploadService.Apprenticeship
+{
+    public class LocationVariationFact
```
gunndabad, line 5: Pascal case property names

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/LocationVariationFacts.cs
```diff
@@ -0,0 +1,558 @@
...
+                    subRegion = "",
+                    expectedError = "Validation error on row 2. Field DELIVERY_METHOD is invalid.",
+                },
+            };
+            foreach (var fact in facts)
+            {
+                Add(fact);
+            }
+
+            /* Test analysis code; uncomment this to generate a csv of all the covered variations which can then be used to check that all desired variations are covered and to facilitate discussions.
```
gunndabad, line 549: If we need this here the `WriteAllText` should work cross-plat (or at least on Windows). `Path.GetTempPath` may be helpful.

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/LocationVariationFact.cs
timabell: Oh yes, whoops. Fixed. Thanks!

Dfc.CourseDirectory.Services.Tests/BulkUploadService/Apprenticeship/LocationVariationFacts.cs
timabell: Done, thanks for the push to improve it.

PR #1596, opened 2020-07-27
[PTCD-508] Set delivery_mode to employer in bulk upload

timabell:
Set delivery_mode to employer if delivery_method is employer. This is the change in behaviour requested in the jira ticket:

> "Current apprenticeship bulk upload implementation will always have delivery modes empty for a location that is employer based. This is different behaviour to the non-bulk upload journey, which sets the delivery modes to Employer. Change required to make bulk upload on apprenticeships to include 'employer' as the delivery mode."

This follows the existing pattern of mutating data in the `Mandatory_Checks_*` methods.

This is the minimum patch needed to achieve the requested behaviour change in order to separate it from refactorings to improve the code.

https://skillsfundingagency.atlassian.net/browse/PTCD-508


timabell:
Passed manual QA with Narahari, see jira comments

PR #1599, opened 2020-07-29
[PTCD-528] Refactor apprenticeship bulk upload controller

timabell:
* Whitespace code cleanup.
* Reduce nesting by using early-return.
* Dispose of stream correctly, adjust test to verify stream during call to mock instead of verify where it is already disposed.

This patch reduces the "cognitive complexity" of the Index action from 31 to 10 as measured by https://plugins.jetbrains.com/plugin/12024-cognitivecomplexity/

There's more we could do on this but we're moving on to other tasks for now.

### Before

upload controller complexity - before

### After

upload controller complexity - after

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Controllers/BulkUploadApprenticeshipsController.cs
+++ b/src/Dfc.CourseDirectory.Web/Controllers/BulkUploadApprenticeshipsController.cs
@@ -78,8 +78,7 @@ namespace Dfc.CourseDirectory.Web.Controllers
             _apprenticeshipService = apprenticeshipService;
         }
 
-        
-        [Authorize]          
+        [Authorize]
         public IActionResult Index()
         {
             _session.SetString("Option", "BulkUploadApprenticeships");
@@ -98,7 +97,7 @@ namespace Dfc.CourseDirectory.Web.Controllers
 
             var courseCounts = _courseService.GetCourseCountsByStatusForUKPRN(new CourseSearchCriteria(UKPRN)).Result;
             var courseErrors = courseCounts.HasValue && courseCounts.IsSuccess ? courseCounts.Value.Where(x => (int)x.Status == (int)RecordStatus.MigrationPending  && x.Count > 0|| (int)x.Status == (int)RecordStatus.MigrationReadyToGoLive && x.Count > 0).Count() : 500;
-            
+
             var model = new BulkUploadViewModel
             {
                 HasMigrationErrors = courseErrors > 0 ? true : false,
@@ -112,7 +111,7 @@ namespace Dfc.CourseDirectory.Web.Controllers
             }
             return View("Index",model);
         }
-         
+
         [Authorize]
         public IActionResult Pending()
         {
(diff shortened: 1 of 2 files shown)


src/Dfc.CourseDirectory.Web/Controllers/BulkUploadApprenticeshipsController.cs
```diff
@@ -131,126 +130,92 @@ public IActionResult Pending()
...
+                }
+                if (!Validate.ValidateFile(bulkUploadFile, out var errorMessage))
+                {
+                    return View(new BulkUploadViewModel {errors = new[] {errorMessage}});
+                }
 
-                    if (!Validate.isBinaryStream(ms))
-                    {
-                        int csvLineCount = _apprenticeshipBulkUploadService.CountCsvLines(ms);
+                await using var ms = new MemoryStream(); // todo: await-using to dispose of stream. Needs test fix.
```
gunndabad, line 145: No need to `Dispose` a `MemoryStream`

timabell: And a todo left behind, oops

tests/Dfc.CourseDirectory.Web.Tests/Controllers/BulkUploadApprenticeshipsControllerTests.cs
```diff
@@ -167,16 +178,12 @@ public async Task IndexPost_Errors()
             VerifyNotUploaded();
         }
 
-        [Matcher]
-        private static Stream ItIsOriginalCsvStream(string originalCsv)
+        private static bool IsOriginalStream(string originalCsv, Stream stream)
```
gunndabad, line 181: I think the name of this method needs reviewing; it's more the contents of the stream that's being compared than the stream itself. Perhaps `MatchesStreamContents` or something?

gunndabad: I don't think this method should be seeking on the stream either.

src/Dfc.CourseDirectory.Web/Controllers/BulkUploadApprenticeshipsController.cs
timabell: According to https://stackoverflow.com/questions/4274590/memorystream-close-or-memorystream-dispose/4274769#4274769 you are correct. I wonder if we should dispose anyway given that it is disposable and that doesn't cause any issues. It would avoid IDE warnings and future discussions at least.

gunndabad: I don't mind keeping the Dispose. It was more a comment on the TODO.

timabell: Ah okay, magic. Todo is about to vanish

timabell: Done

tests/Dfc.CourseDirectory.Web.Tests/Controllers/BulkUploadApprenticeshipsControllerTests.cs
timabell: Pushed a new naming, what do you think?

gunndabad: I like revised name

timabell: Also pushed a commit to make the controller responsible for the memory stream position, rewinding it each time it passes it to other things. What do you think of that commit?

PR #1601, opened 2020-07-30
Add new CsvUtil class with a CountLines method

gunndabad:


src/Dfc.CourseDirectory.Core/CsvUtil.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.IO;
+
+namespace Dfc.CourseDirectory.Core
+{
+    public static class CsvUtil
```
timabell, line 5: Classes called util attract other stuff, how about:
```suggestion
    public static class CsvLineCounter
```

gunndabad: I have another change Coming Soon (TM) which adds something else here so the naming will make more sense then.

PR #1603, opened 2020-07-31
Add new IBinaryStorageProvider abstraction and an implementation using Blob Storage

gunndabad:


src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Dfc.CourseDirectory.WebV2.BinaryStorageProvider
+{
+    public class BlobStorageBinaryStorageProvider : IBinaryStorageProvider
+    {
+        private readonly CloudBlobClient _blobClient;
+        private readonly string _containerName;
+
+        public BlobStorageBinaryStorageProvider(
+            string accountName,
```
jameslawesynergy, line 15: I'm leaning these days towards passing primitive types in a container class or using the `Options` pattern.

src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+            string accountKey,
+            string containerName)
+        {
+            var account = new CloudStorageAccount(new StorageCredentials(accountName, accountKey), useHttps: true);
+            _blobClient = account.CreateCloudBlobClient();
+
+            _containerName = containerName;
```
jameslawesynergy, line 22: Do you need to call `CreateContainerIfNotExists` here?

src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+        public async Task<bool> TryDownloadFile(string path, Stream destination)
+        {
+            var blobRef = _blobClient.GetContainerReference(_containerName).GetBlockBlobReference(path);
+
+            if (!await blobRef.ExistsAsync())
+            {
+                return false;
+            }
+
+            await blobRef.DownloadToStreamAsync(destination);
```
jameslawesynergy, line 34: Worth perhaps checking the result to be safe on the potential race condition?

src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
gunndabad: I have another patch that uses Options here but want to go over some config-related stuff in tech sessions first.

src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
gunndabad: What's the race?

jameslawesynergy: Blob could be deleted in between `Exists()` check and `Download()`.

Suggest `try { ... } catch (StorageException ex) when (ex.StatusCode == 412) { return false; }` (Don't quote me on the exception and code type.)

gunndabad: Nothing ever deletes blobs here

src/Dfc.CourseDirectory.Core/Dfc.CourseDirectory.Core.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="HtmlAgilityPack" Version="1.11.22" />
     <PackageReference Include="HtmlSanitizer" Version="5.0.274-beta" />
     <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="2.4.0" />
+    <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="10.0.3" />
```
jameslawesynergy, line 15: Consider using `Azure.Storage.Blobs` library (I believe this is the latest/recommended API to use)

src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
jameslawesynergy: That's no guarantee that a different API/consumer of this data won't in the future.

src/Dfc.CourseDirectory.WebV2/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -35,11 +36,19 @@ public async Task<bool> TryDownloadFile(string path, Stream destination)
...
         {
-            var blobRef = _blobClient.GetContainerReference(_containerName).GetBlockBlobReference(path);
+            var container = await GetContainer();
+            var blobRef = container.GetBlockBlobReference(path);
 
-            return blobRef.UploadFromStreamAsync(source);
+            await blobRef.UploadFromStreamAsync(source);
+        }
+
+        private async Task<CloudBlobContainer> GetContainer()
```
jameslawesynergy, line 47: I believe it's possible to hold the reference to the BlobContainer as a private property (as it's just a client under the hood) to avoid having to execute this codepath an excessive number of times.

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+            var blob = _blobContainerClient.GetBlobClient(path);
+            await blob.UploadAsync(source);
+        }
+
+        private async Task EnsureContainerExists()
+        {
+            if (!_containerExists)
+            {
+                await _blobContainerClient.CreateIfNotExistsAsync();
```
jameslawesynergy, line 48: You could move this to ctor to avoid executing it every time. (There's a non-asyn overload for it)

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+            await EnsureContainerExists();
+
+            var blob = _blobContainerClient.GetBlobClient(path);
+
+            try
+            {
+                await blob.DownloadToAsync(destination);
+                return true;
+            }
+            catch (RequestFailedException ex) when (ex.Status == 404)
```
jameslawesynergy, line 30: I prefer to use `StatusCodes.Status404NotFound` for readability.

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public async Task<bool> TryDownloadFile(string path, Stream destination)
+        {
+            await EnsureContainerExists();
+
+            var blob = _blobContainerClient.GetBlobClient(path);
+
+            try
+            {
+                await blob.DownloadToAsync(destination);
```
jameslawesynergy, line 27: I'm not sure of your intended consumption of this method, however it might be more flexible if you were to return the clients stream directly rather than copying onto a new stream...

```
public async Task<(bool success, Stream result)> Download(string blobName)
{
    var blob = _blobContainerClient.GetBlobClient(path);

    try
    {
        return (true, (await blobClient.DownloadAsync()).Value.Content);
    }
    catch (RequestFailedException ex)
        when (ex.Status == StatusCodes.Status404NotFound)
    {
        return (false, null);
    }
}
```

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        private readonly BlobContainerClient _blobContainerClient;
+        private bool _containerExists;
+
+        public BlobStorageBinaryStorageProvider(IOptions<BlobStorageBinaryStorageProviderSettings> options)
+        {
+            _blobContainerClient = new BlobContainerClient(options.Value.ConnectionString, options.Value.ContainerName);
+        }
+
+        public async Task<bool> TryDownloadFile(string path, Stream destination)
```
jameslawesynergy, line 19: Perhaps check for `string.IsNullOrWhiteSpace(path)`?

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
```diff
@@ -0,0 +1,53 @@
...
+                await blob.DownloadToAsync(destination);
+                return true;
+            }
+            catch (RequestFailedException ex) when (ex.Status == 404)
+            {
+                return false;
+            }
+        }
+
+        public async Task UploadFile(string path, Stream source)
```
jameslawesynergy, line 36: Perhaps check `path `and `stream` for null?

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
gunndabad: That's an ASP.NET Core type - don't want to introduce a dependency on that in this project

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
gunndabad: I'm rather keep the constructor dump and avoid doing IO there if possible

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
gunndabad: My thinking behind current signature was you could stream a blob directly to an `HttpResponse`'s body without ever having the whole thing in memory.

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
jameslawesynergy: You're quite right, I though it was standard! Could use `System.Net.HttpStatusCode.NotFound` then?

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
jameslawesynergy: If that's your use case then I'd certainly suggest using `DownloadAync()` and grabbing `Value.Content`. With your original approach you're going to need to new-up a `MemoryStream` to pass as a parameter to the method which will involve writing it to memory.

jameslawesynergy: NB the value tuple in my suggestion isn't a great idea! :)

gunndabad: Why would you need a `MemoryStream` at all? You can write directly to `HttpResponse.Body` (or do it via an `IActionResult`)

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
jameslawesynergy: That's fair enough. Although you have a potential concurrency problem with the current implementation as you'll need to ensure that only a single thread attempts to call` CreateIfNotExistsAsync()`. Suggest locking with `SemaphoreSlim`.

gunndabad: Pretty sure that `CreateIfNotExistsAsync` is thread-safe. I see a benign race but can't see how it would cause any problems.

src/Dfc.CourseDirectory.Core/BinaryStorageProvider/BlobStorageBinaryStorageProvider.cs
jameslawesynergy: When calling `TryDownloadFile(string path, Stream destination)` you need a pass an initialised Stream object which has to come from somewhere. (Please correct me if I'm wrong, but I don't believe `HttpResponseMessage `or any of the `IActionResult `implementations provide you with an initialised `Stream` object to pass.)

PR #1604, opened 2020-08-03
Sync Courses to SQL (PTCD-377)

gunndabad:


src/Dfc.CourseDirectory.Core/DataStore/Sql/Queries/UpsertCourses.cs
```diff
@@ -0,0 +1,58 @@
...
+        public string EntryRequirements { get; set; }
+        public string WhatYoullLearn { get; set; }
+        public string HowYoullLearn { get; set; }
+        public string WhatYoullNeed { get; set; }
+        public string HowYoullBeAssessed { get; set; }
+        public string WhereNext { get; set; }
+        public IEnumerable<UpsertCoursesRecordCourseRun> CourseRuns { get; set; }
+    }
+
+    public class UpsertCoursesRecordCourseRun
```
timabell, line 34: I prefer class-per-file, gives me one less thing to think about

gunndabad: This is consistent with other SQL-sync commands. Happy to review the approach but I'd rather keep consistent until then.

timabell: Agreed

PR #1620, opened 2020-08-13
PTCD-419 Added Published confirmation page to AddCourse

jameslawesynergy:
Added Published confirmation page to AddCourse

image

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Controllers/AddCourseController.cs
+++ b/src/Dfc.CourseDirectory.Web/Controllers/AddCourseController.cs
@@ -1,14 +1,11 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Collections.Immutable;
-using System.Globalization;
 using System.Linq;
 using System.Text.Encodings.Web;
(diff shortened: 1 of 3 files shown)


src/Dfc.CourseDirectory.Web/Controllers/AddCourseController.cs
```diff
@@ -883,12 +880,14 @@ public async Task<IActionResult> AcceptAndPublish()
 
             if (result.IsSuccess && result.HasValue)
             {
-                return RedirectToAction("Index", "ProviderCourses",
-                    new {
-                        CourseRunId = courseRuns[0].id,
-                        NotificationTitle = "Course added",
-                        NotificationMessage = "You added",
-                    });
+                Session.SetObject(SessionPublishedCourse, new PublishedCourseViewModel
```
jameslawesynergy, line 883: Ok, so this is a pretty dirty abuse of the session state to pass the data to the next page, but it means that we can ensure the confirmation is only displayed once if the user clicks refresh then it doesn't attempt to publish the course again.

PR #1628, opened 2020-08-19
Delete duplicate file

gunndabad:
Duplicates `src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/GetProviderUKPRNForApprenticeship.cs`


timabell (approved):
Mentioning in the PR the path of the other copy would be handy

PR #1635, opened 2020-08-24
PTCD-650 Update Edit Course Journey

jameslawesynergy:
Updated NotificationTagHelper to add Success variant and remove heading margin when content is empty
Updated EditCourseRun page and journey
Updated EditCourse page and journey

image
image
image


src/Dfc.CourseDirectory.WebV2/Behaviors/TempDataKeys.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Dfc.CourseDirectory.WebV2.Behaviors
```
gunndabad, line 1: This namespace is more for MediatR cross-cutting stuff. Could go in root `WebV2` namespace, for me.

gunndabad (approved):

PR #1641, opened 2020-08-28
Rationalise Provider Context

gunndabad:
Provider Context is conceptually the 'current provider'. The legacy views use the `UKPRN` in Session for storing this; v2 bits can use a `providerId` query parameter, a `providerId` route value or the provider ID can be extracted from whatever entity is being acted upon (e.g. a Course).

This PR removes some of the inconsistencies and ties up some lose ends around the current usage. A separate PR will follow that removes the `VerifyApprenticeshipId` attributes & filters and replaces it with something in line with what we do for courses (i.e. a MediatR behavior).

- Use IProviderContextProvider everywhere for getting/setting provider context.
- Move the session-based fallback into IProviderContextProvider
- Remove IRequiresProviderContextController (it was a convenience that brings more confusion than good)
- Set the provider context when authorizing access to a course. (Guarantees that the provider context set matches the provider for the course).

### DIFF
--- a/src/Dfc.CourseDirectory.WebV2/Behaviors/RequireUserCanAccessCourseBehavior.cs
+++ b/src/Dfc.CourseDirectory.WebV2/Behaviors/RequireUserCanAccessCourseBehavior.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Threading;
+﻿using System.Threading;
 using System.Threading.Tasks;
 using Dfc.CourseDirectory.Core;
 using Dfc.CourseDirectory.WebV2.Security;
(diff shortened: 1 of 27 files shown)


src/Dfc.CourseDirectory.WebV2/Behaviors/RequireUserCanAccessCourseBehavior.cs
```diff
@@ -34,25 +39,42 @@ public async Task<TResponse> Handle(TRequest request, CancellationToken cancella
                 throw new ResourceDoesNotExistException(ResourceType.Course, courseId);
             }
 
-            var currentUser = _currentUserProvider.GetCurrentUser();
+            var isAuthorized = IsAuthorized();
 
-            if (currentUser == null)
+            if (isAuthorized)
```
timabell, line 44: I'd invert the if to reduce nesting

timabell (approved):

soharadesign (approved):

PR #1643, opened 2020-08-28
Fixed CourseSummary subregions displaying as regions bug

jameslawesynergy:
Fixed CourseSummary subregions displaying as regions bug

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
+++ b/src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
@@ -1,13 +1,8 @@
 ﻿using System;
-using System.Collections.Generic;
-
 using System.Linq;
-using System.Reflection;
-using System.Threading.Tasks;
 using Dfc.CourseDirectory.Common;
 using Dfc.CourseDirectory.Models.Interfaces.Courses;
 using Dfc.CourseDirectory.Models.Models.Courses;
-using Dfc.CourseDirectory.Models.Models.Regions;
 using Dfc.CourseDirectory.Services.CourseService;
 using Dfc.CourseDirectory.Services.Interfaces.CourseService;
 using Dfc.CourseDirectory.Services.Interfaces.VenueService;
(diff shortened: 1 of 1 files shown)


src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
```diff
@@ -118,45 +118,20 @@ public IActionResult Index(Guid? courseId, Guid? courseRunId)
...
-                var regions = GetRegions().RegionItems.Select(x => x.Id);
-                vm.Regions = FormattedRegionsByIds(allRegions, courseRun.Regions);
+                var allRegions = _courseService.GetRegions();
+
+                vm.Regions = allRegions.RegionItems
+                    .Select(r => new { r.Id, r.RegionName })
+                    .Concat(allRegions.RegionItems
+                        .SelectMany(r => r.SubRegion)
+                        .Select(r => new { r.Id, RegionName = r.SubRegionName }))
+                    .Where(r => allRegions.SubRegionsDataCleanse(courseRun.Regions.ToList()).Contains(r.Id))
```
timabell, line 130: I think that `ToList` going to be run for every record. You might want to extract it to a variable or something.

src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
```diff
@@ -118,45 +118,20 @@ public IActionResult Index(Guid? courseId, Guid? courseRunId)
...
 
-            if(courseRun.Regions != null)
+            if (courseRun.Regions != null)
             {
-                var allRegions = _courseService.GetRegions().RegionItems;
-                var regions = GetRegions().RegionItems.Select(x => x.Id);
-                vm.Regions = FormattedRegionsByIds(allRegions, courseRun.Regions);
+                var allRegions = _courseService.GetRegions();
+
+                vm.Regions = allRegions.RegionItems
```
timabell, line 125: I'm finding this linq pretty hard to parse in terms of intent. Could you figure out a way to make the intent clearer?

timabell (approved):
[I would expect to see addition of test coverage for modifying legacy code.](https://timwise.co.uk/2020/07/09/approaches-to-refactoring-and-technical-debt/)

src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
jameslawesynergy: Good spot - fixed.

src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
jameslawesynergy: Updated as discussed.

jameslawesynergy:
Cathie to confirm with Alix/Danny if this bug requires fixing or not.

jameslawesynergy:
Discussed with Cathie:

re courses and regions v sub-regions having different displays, discussed with Danny, Alix and Clair - this is prior behaviour so they haven't changed anything.  UR would be best for this and we don't have this yet.  There could be some providers that have lots of sub-regions and we don't know how this would look.
tl:dr; please leave courses as-is for the mo

PR #1644, opened 2020-08-28
PTCD-493 Update Copy CourseRun Journey

jameslawesynergy:
Add Check and Publish and Published pages to CopyCourseRun journey

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
+++ b/src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
@@ -122,9 +122,9 @@ namespace Dfc.CourseDirectory.Web.Controllers.CopyCourse
 
             var regions = _courseService.GetRegions();
 
-            foreach (var subRegion in regions?.RegionItems.SelectMany(r => r.SubRegion))
+            foreach (var subRegion in regions.RegionItems.SelectMany(r => r.SubRegion))
             {
-                subRegion.Checked = courseRun.Regions.Contains(subRegion.Id);
+                subRegion.Checked = courseRun.Regions?.Contains(subRegion.Id);
             }
 
             CopyCourseRunViewModel vm = new CopyCourseRunViewModel
@@ -247,29 +247,29 @@ namespace Dfc.CourseDirectory.Web.Controllers.CopyCourse
             }
             else
             {
-                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+                return NotFound();
             }
 
             var course = await _courseService.GetCourseByIdAsync(new GetCourseByIdCriteria(courseId.Value));
 
             if (!course.HasValue)
             {
-                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+                return NotFound();
             }
 
             var courseRun = course.Value.CourseRuns.SingleOrDefault(cr => cr.id == courseRunId);
 
             if (courseRun == null)
             {
-                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+                return NotFound();
             }
 
             var venues = await _venueService.SearchAsync(new VenueSearchCriteria(ukprn.ToString(), null));
             var regions = _courseService.GetRegions();
 
-            foreach (var subRegion in regions?.RegionItems.SelectMany(r => r.SubRegion))
+            foreach (var subRegion in regions.RegionItems.SelectMany(r => r.SubRegion))
             {
-                subRegion.Checked = courseRun.Regions.Contains(subRegion.Id);
+                subRegion.Checked = courseRun.Regions?.Contains(subRegion.Id);
             }
 
             var vm = new CopyCourseRunViewModel
@@ -339,9 +339,9 @@ namespace Dfc.CourseDirectory.Web.Controllers.CopyCourse
 
                 if (savedModel.DeliveryMode == DeliveryMode.WorkBased && !savedModel.National)
                 {
-                    foreach (var subRegion in regions?.RegionItems.SelectMany(r => r.SubRegion))
+                    foreach (var subRegion in regions.RegionItems.SelectMany(r => r.SubRegion))
                     {
-                        subRegion.Checked = savedModel.SelectedRegions.Contains(subRegion.Id);
+                        subRegion.Checked = savedModel.SelectedRegions?.Contains(subRegion.Id);
                     }
                 }
             }
@@ -359,7 +359,7 @@ namespace Dfc.CourseDirectory.Web.Controllers.CopyCourse
 
             if (model == null)
             {
-                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+                return NotFound();
             }
 
             _session.SetObject(CopyCourseRunSaveViewModelSessionKey, model);
@@ -381,7 +381,7 @@ namespace Dfc.CourseDirectory.Web.Controllers.CopyCourse
 
             if (model == null)
             {
-                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+                return NotFound();
             }
 
             var availableRegions = new Lazy<SelectRegionModel>(() => _courseService.GetRegions());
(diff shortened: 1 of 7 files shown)


src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
-                courseRunVenues.Add(item);
-            };
+                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+            }
 
+            var venues = await _venueService.SearchAsync(new VenueSearchCriteria(ukprn.ToString(), null));
             var regions = _courseService.GetRegions();
 
-            if (courseId.HasValue)
+            foreach (var subRegion in regions?.RegionItems.SelectMany(r => r.SubRegion))
```
gunndabad, line 278: Could NullRef here if `regions` is `null`

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
+            if (courseId.HasValue && courseRunId.HasValue)
+            {
+                if (savedModel != null)
+                {
+                    if (savedModel.CourseId != courseId || savedModel.CourseRunId != courseRunId)
+                    {
+                        _session.Remove(CopyCourseRunSaveViewModelSessionKey);
+                        savedModel = null;
+                    }
+                    else
```
gunndabad, line 244: Isn't this `else` block redundant?

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
+            else if (savedModel != null)
             {
-                UKPRN = _session.GetInt32("UKPRN").Value;
+                courseId = savedModel.CourseId;
+                courseRunId = savedModel.CourseRunId;
             }
             else
             {
-                return RedirectToAction("Index", "Home", new { errmsg = "Please select a Provider." });
+                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
```
gunndabad, line 258: Could perhaps return a 404 here instead..

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
             }
 
-            List<SelectListItem> courseRunVenues = new List<SelectListItem>();
+            var course = await _courseService.GetCourseByIdAsync(new GetCourseByIdCriteria(courseId.Value));
 
-            VenueSearchCriteria criteria = new VenueSearchCriteria(UKPRN.ToString(), null);
-            var venues = await _venueService.SearchAsync(criteria);
+            if (!course.HasValue)
+            {
+                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
```
gunndabad, line 265: As above.. 404?

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
 
-            foreach (var venue in venues.Value.Value)
+            var courseRun = course.Value.CourseRuns.SingleOrDefault(cr => cr.id == courseRunId);
+
+            if (courseRun == null)
             {
-                var item = new SelectListItem { Text = venue.VenueName, Value = venue.ID };
-                courseRunVenues.Add(item);
-            };
+                return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
```
gunndabad, line 272: Ditto - 404?

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
-                            }
-                        }
+                        venue.Selected = venue.Value == savedModel.VenueId.ToString();
                     }
+                }
 
-                    if (vm.ChooseRegion.Regions.RegionItems != null && vm.ChooseRegion.Regions.RegionItems.Any())
+                if (savedModel.DeliveryMode == DeliveryMode.WorkBased && !savedModel.National)
+                {
+                    foreach (var subRegion in regions?.RegionItems.SelectMany(r => r.SubRegion))
```
gunndabad, line 350: As above - could throw NRE if `regions` is `null`

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
```diff
@@ -262,243 +219,355 @@ private async Task<SelectVenueModel> GetVenuesByUkprn(int ukprn)
...
-            return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
+            course.Value.CourseRuns = course.Value.CourseRuns.Append(courseRun);
+
+            try
+            {
+                var result = await _courseService.UpdateCourseAsync(course.Value);
+
+                if (!result.IsSuccess || !result.HasValue)
+                {
+                    return View("Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
```
gunndabad, line 538: I think this should throw otherwise we won't see the error in logs

gunndabad (approved):

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: No, as you only want to execute that code when the previous condition fails. These conditions are quite confusing so I'm going to invert this if to make it clearer.

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: `Regions` should never be `null`, so the use of the null conditional operator here is misleading.

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: `Regions` should never be `null`, so the use of the null conditional operator here is misleading.

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: I've not actually changed the behaviour of this from the original (just inverted ifs to reduce nesting)

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: Agreed.

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: Agreed.

src/Dfc.CourseDirectory.Web/Controllers/CopyCourse/CopyCourseRunController.cs
jameslawesynergy: Agreed.

PR #1655, opened 2020-09-04
Fix flaky test

gunndabad:


timabell:
From a quick read I don't really understand what this is doing and why the removed line of code is no longer wanted. I'd suggest amending the commit to explain why it's okay to kill it, and maybe a copy of the error message we were getting when it flapped.

tests/Dfc.CourseDirectory.Core.Tests/BackgroundWorkersTests/QueueBackgroundWorkSchedulerTests.cs
```diff
@@ -125,7 +125,6 @@ public async Task PassesCancellationTokenToWorkItemThatIsCancelledWhenServiceSto
             await scheduler.StartAsync(cancellationToken: default);
             await scheduler.Schedule(workItem);
             executingSignal.Wait(100);
-            Assert.NotEqual(default, itemCancellationToken);
```
jameslawesynergy, line 128: I think there's something else weird going on here. `default(CancellationToken)` should resolve to `CancellationToken.None` (https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken.none?redirectedfrom=MSDN&view=netcore-3.1#System_Threading_CancellationToken_None), however I would not expect that method to ever receive `CancellationToken.None` as a parameter unless there's something wonky going on under the hood.

Perhaps make the declaration nullable (`112. CancellationToken? itemCancellationToken = null;`) and then assert `itemCancellationToken.HasValue`?

gunndabad:
Test seems to no longer be flaky

PR #1662, opened 2020-09-10
PTCD-670 Add PurgeLarsDataFromDatabaseAndSearch Function

jameslawesynergy:


src/Dfc.CourseDirectory.Functions/local.settings.json
```diff
@@ -3,5 +3,8 @@
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
     "FUNCTIONS_WORKER_RUNTIME": "dotnet"
+  },
+  "Host": {
```
jameslawesynergy, line 7: I've added this so that manual trigger functions can be executed via the admin HTTP endpoint (`POST http://localhost:7071/admin/functions/{FunctionName}`) without having to reconfigure it every time.

src/Dfc.CourseDirectory.Core/DataStore/Sql/Queries/DeleteLarsLearningDelivery.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using OneOf;
+using OneOf.Types;
+
+namespace Dfc.CourseDirectory.Core.DataStore.Sql.Queries
+{
+    public class DeleteLarsLearningDelivery : ISqlQuery<OneOf<string, None>>
```
gunndabad, line 7: `ISqlQuery<OneOf<NotFound, Success>>`?

src/Dfc.CourseDirectory.Core/DataStore/Sql/Queries/DeleteLarsLearningDelivery.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public string LearnAimRef { get; }
```
gunndabad, line 9: We tend to use POCOs here instead of read-only properties and a non-default constructor (bring on C# 9 records)

src/Dfc.CourseDirectory.Functions/Config/SearchClientSettings.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace Dfc.CourseDirectory.Functions.Config
+{
+    public class SearchClientSettings
```
gunndabad, line 5: As above Settings classes are usually POCOs with a default constructor

Resources/Azure/ArmTemplates/template.json
```diff
@@ -722,6 +722,14 @@
               {
                 "name": "CosmosDbSettings__ConnectionString",
                 "value": "[parameters('providerPortalCosmosDbConnectionString')]"
+              },
+              {
```
gunndabad, line 726: Suggest adding a prefix to these like the others so we can easily bind an `IOptions<T>` from a config section later if required

gunndabad (approved):

src/Dfc.CourseDirectory.Core/DataStore/Sql/Queries/DeleteLarsLearningDelivery.cs
jameslawesynergy: Agreed

Resources/Azure/ArmTemplates/template.json
jameslawesynergy: I think we should cross that bridge when we come to it as these values aren't exposed any further than this application at this time.

PR #1666, opened 2020-09-15
PTCD-470 Remove non-alphanumeric characters from LARS search

jameslawesynergy:


src/Dfc.CourseDirectory.Web/Helpers/LarsSearchHelper.cs
```diff
@@ -260,13 +260,12 @@ internal static string FormatSearchTerm(string searchTerm)
...
             Throw.IfNullOrWhiteSpace(searchTerm, nameof(searchTerm));
 
-            var split = searchTerm
-                .Split(' ')
-                .Select(x => x.Trim())
-                .Where(x => !string.IsNullOrWhiteSpace(x))
-                .ToList();
+            var segments = new string(searchTerm.Where(c => char.IsLetterOrDigit(c) || char.IsWhiteSpace(c)).ToArray())
+                .Split(" ")
+                .Where(s => !string.IsNullOrWhiteSpace(s))
```
gunndabad, line 265:
```suggestion
                .Split(" ", StringSplitOptions.RemoveEmptyEntries)
```

gunndabad (approved):

src/Dfc.CourseDirectory.Web/Helpers/LarsSearchHelper.cs
jameslawesynergy: `StringSplitOptions.RemoveEmptyEntries` doesn't consider whitespace characters as empty - it's an unlikely scenario that whitespace chars other than space would find their way into the search query, but not beyond the realms of impossible, so we would still require the `Where(s => !string.IsNullOrWhiteSpace(s))` in there.

PR #1672, opened 2020-09-21
Add an implementation of ISession{Store} for testing

gunndabad:
Use a single dictionary to back all sessions. This allows us to
query/amend the session data outside of an HTTP request.


timabell (approved):
I think you should copy the description of this PR into xml doc for the new class `SingletonSession`

soharadesign (approved):

PR #1683, opened 2020-10-02
Ncsd 4352 dev ops upgrade

trivedihimanshu:


olusola-adio (approved):

.github/CODEOWNERS
```diff
@@ -1 +1,3 @@
-* @gunndabad @javeedESyn @MrKevJoy @soharadesign @timabell @jameslawesynergy
+#Requires review from DevOps when making changes to any files in the Resources folder or it's subfolders
+/Resources/    @SkillsFundingAgency/dfc-devops
```
timabell, line 2: This repository must be fully owned by the team that actually works on it.

timabell (changes requested):

.github/CODEOWNERS
trivedihimanshu: Resources folder and all the repositories are managed/administrated by DevOps team.

Project team have write access within the repo and code developed by Developers/Testers.

I don't think you have understood the changes I've made within Codeowner file.

timabell: You could be right @trivedihimanshu, I'm not that familiar with the `CODEOWNERS` system; I've never been on a team that's had any need for it, people just respect each other's areas without formal control mechanisms.

My assumption was that this patch will **require** an approval from the "devops team" (who are external to the development team building and running this service) for any changes to the `/Resources/` folder. It might be that I'm wrong and that it's just a notification setting.

From reading the docs I'm still not 100% clear how this actually behaves.
https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners

If you could clarify the intent of this change in your commit message and PR message that would be helpful for reviewing it.

I'll leave it rejected for now to make sure we get a bit of time to understand it properly before it's merged in.

Fundamentally I believe that development teams should be responsible for operating their live service to avoid the "Dev vs Ops" ping-pong that regularly happens in orgs that have not truly embraced DevOps. For me the term DevOps means - "Developers doing Operations"; it's a matter of keeping responsibility and control within the same team. This doesn't mean developers flailing around outside their area of expertise with out help, in fact it's important that the development team has the support of specialists in infrastructure etc, either in the form of properly embedded infrastructure engineers, or people providing expertise to the team.

trivedihimanshu: @timabell  I don’t understand your concern still

Resources folder and sub-folders under Resources contains IaC which owned and managed by DevOps (edited) 

Also i don’t need to get in to arguments or discussion in which we would simply argue for hours about correct definition and understanding of DevOps, i’ve applied standard approach which is followed and adhere by everyone who works for DFE @soharadesign  FYI

timabell: Okay, if you don't wish to discuss further then I'll defer to @soharadesign as tech lead, and he can decide whether to allow this through.

In its current state personally I would not let this PR be merged.

trivedihimanshu: I never requested you to review this PR as I was working with Steve on this and I was expecting him to review the PR.

soharadesign (approved):

.github/CODEOWNERS
soharadesign: Happy to approve as this is consistent with the rest of the DfE. If this causes issues down the line, we can raise it for further discussion.

PR #1696, opened 2020-10-16
Remove Frameworks from apprenticeship search results (PTCD-631)

gunndabad:
Also simplified the `FindStandard` journey.

Legacy code is left as-is as it is not easy to port to the v2 version.


src/Dfc.CourseDirectory.Web/Controllers/ApprenticeshipsController.cs
```diff
@@ -102,7 +102,7 @@ public async Task<IActionResult> Search([FromQuery] SearchRequestModel requestMo
 
                 if (result.IsSuccess && result.HasValue)
                 {
-                    foreach (var item in result.Value)
+                    foreach (var item in result.Value.Where(r => r.ApprenticeshipType != ApprenticeshipType.FrameworkCode))
```
timabell, line 105: Wondering out loud if we should whitelist rather than blacklist here. On the plus side it would mean there's no mention here of a concept from the past (Frameworks).

src/Dfc.CourseDirectory.WebV2/Features/Apprenticeships/FindStandard.cs
```diff
@@ -43,49 +33,33 @@ public class ViewModel
...
     public class Handler :
         IRequestHandler<Query, ViewModel>,
         IRestrictProviderType<Query>,
         IRequestHandler<SearchQuery, QueryResponse>,
-        IRestrictProviderType<SearchQuery>,
-        IRequestHandler<SelectCommand, Success>,
-        IRestrictProviderType<SelectCommand>
+        IRestrictProviderType<SearchQuery>
     {
         private readonly IStandardsAndFrameworksCache _standardsAndFrameworksCache;
```
timabell, line 52: Can we rename/update `IStandardsAndFrameworksCache` yet or is there other stuff depending on it? Or is that another PR?

timabell (approved):
Might be worth a rebase before merging, it's based a few PRs behind `main`.

Negative patch size. Win.

Good stuff, shipit :ship:

src/Dfc.CourseDirectory.WebV2/Features/Apprenticeships/FindStandard.cs
gunndabad: Yeh there's still a few things that refer to it (e.g. bulk upload)

PR #1698, opened 2020-10-16
Editorconfig - add final newline

timabell:
Editorconfig - add final newline

Because git and github complain if you don't which is minorly annoying.


gunndabad (approved):

timabell:
rebased

PR #1715, opened 2020-10-29
Use new search abstractions for ONSPD lookup

gunndabad:


src/Dfc.CourseDirectory.Web/Helpers/OnspdSearchHelper.cs
```diff
@@ -1,28 +1,30 @@
...
-        public OnspdSearchHelper(IOnspdService onspdService)
+        public OnspdSearchHelper(ISearchClient<Core.Search.Models.Onspd> searchClient)
         {
-            _onspdService = onspdService;
+            _searchClient = searchClient;
         }
 
         public Onspd GetOnsPostcodeData(string postcode)
         {
             var onspd = new Onspd();
```
gunndabad, line 19: Returning an empty `Onspd` object here if the search failed here is fishy but keeping as-is for now

src/Dfc.CourseDirectory.Web/Helpers/OnspdSearchHelper.cs
```diff
@@ -1,28 +1,30 @@
...
             if (!string.IsNullOrWhiteSpace(postcode))
             {
-                var onspdSearchCriteria = new OnspdSearchCriteria(postcode);
-                var onspdResult = _onspdService.GetOnspdData(onspdSearchCriteria);
-                if (onspdResult.IsSuccess && onspdResult.HasValue)
+                var searchResult = _searchClient.Search(new OnspdSearchQuery() { Postcode = postcode })
```
gunndabad, line 22: Sync-over-async to maintain current method signature

Resources/ArmTemplates/template.json
```diff
@@ -473,22 +473,6 @@
...
-              {
-                "name": "OnspdSearchSettings__SearchServiceAdminApiKey",
-                "value": "[parameters('searchAdminKey')]"
-              },
-              {
-                "name": "OnspdSearchSettings__SearchServiceQueryApiKey",
-                "value": "[parameters('searchQueryKey')]"
-              },
-              {
-                "name": "OnspdSearchSettings__IndexName",
```
jameslawesynergy, line 489: Is it worth keeping this configurable if we need to change indexes over in future?

src/Dfc.CourseDirectory.Web/Helpers/OnspdSearchHelper.cs
```diff
@@ -1,28 +1,30 @@
...
+                    .GetAwaiter().GetResult();
+
+                if (searchResult.Results.Count > 0)
                 {
-                    onspd = onspdResult.Value.Value;
+                    return searchResult.Results.Single().Adapt<Onspd>();
```
jameslawesynergy, line 27: I don't trust AzureSearch to only return one record. Perhaps safer to use `First()`?

Although this could be expressed more easily with `return searchResults.Results.FirstOrDefault()?.Adapt<Onspd>() ?? new Onspd()`? (Would remove nesting and the `onspd` var)

src/Dfc.CourseDirectory.Web/Helpers/OnspdSearchHelper.cs
```diff
@@ -1,28 +1,30 @@
...
+        private readonly ISearchClient<Core.Search.Models.Onspd> _searchClient;
 
-        public OnspdSearchHelper(IOnspdService onspdService)
+        public OnspdSearchHelper(ISearchClient<Core.Search.Models.Onspd> searchClient)
         {
-            _onspdService = onspdService;
+            _searchClient = searchClient;
         }
 
         public Onspd GetOnsPostcodeData(string postcode)
```
jameslawesynergy, line 17: This is only called 3 times from one class, you could trash this unnecessary abstraction completely?

jameslawesynergy (approved):

src/Dfc.CourseDirectory.Web/Helpers/OnspdSearchHelper.cs
gunndabad: The query has a `Size` == `1`

Resources/ArmTemplates/template.json
gunndabad: There's a separate (existing) config entry for the index name

PR #1717, opened 2020-11-02
PTCD-696 Suppress Apprenticeship Frameworks

jameslawesynergy:
Updated ApprenticeshipsController.ViewAndEdit to show only found Apprenticeships with RecordStaus.Live
Added GetApprenticeships Cosmos query and GetAllFrameworkApprenticeshipIds Function
Added ArchiveApprenticeships Function

### DIFF
--- a/src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/GetApprenticeships.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/GetApprenticeships.cs
@@ -1,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Models;
+
+namespace Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries
+{
+    public class GetApprenticeships : ICosmosDbQuery<IDictionary<Guid, Apprenticeship>>
+    {
+        public GetApprenticeships(Expression<Func<Apprenticeship, bool>> predicate = null)
+        {
+            Predicate = predicate;
+        }
+
+        public Expression<Func<Apprenticeship, bool>> Predicate { get; set; }
+    }
+}
--- a/src/Dfc.CourseDirectory.Functions/GetAllPublishedFrameworkApprenticeshipIds.cs
+++ b/src/Dfc.CourseDirectory.Functions/GetAllPublishedFrameworkApprenticeshipIds.cs
@@ -1,0 +1,31 @@
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.Extensions.Logging;
+
+namespace Dfc.CourseDirectory.Functions
+{
+    public class GetAllPublishedFrameworkApprenticeshipIds
+    {
+        private readonly ICosmosDbQueryDispatcher _cosmosDbQueryDispatcher;
+
+        public GetAllPublishedFrameworkApprenticeshipIds(ICosmosDbQueryDispatcher cosmosDbQueryDispatcher)
+        {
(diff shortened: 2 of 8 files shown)


src/Dfc.CourseDirectory.Functions/GetAllPublishedFrameworkApprenticeshipIds.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public GetAllPublishedFrameworkApprenticeshipIds(ICosmosDbQueryDispatcher cosmosDbQueryDispatcher)
+        {
+            _cosmosDbQueryDispatcher = cosmosDbQueryDispatcher ?? throw new ArgumentNullException(nameof(cosmosDbQueryDispatcher));
+        }
+
+        [FunctionName("GetAllPublishedFrameworkApprenticeshipIds")]
+        public async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get")] HttpRequest req, ILogger log)
+        {
+            var results = await _cosmosDbQueryDispatcher.ExecuteQuery(new GetApprenticeships(a => a.FrameworkId != null));
```
gunndabad, line 26: Need an extra condition in the predicate here to check status (or the query is mis-named); `GetApprenticeships` doesn't look to be only returning published records

jameslawesynergy: https://github.com/SkillsFundingAgency/dfc-coursedirectory/pull/1717/files#diff-4ecd5d0b7c870949dc17394bc2c3e3efad0b75be40cf3fe318098835a0ceef51R21?

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/GetApprenticeships.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Models;
+
+namespace Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries
+{
+    public class GetApprenticeships : ICosmosDbQuery<IDictionary<Guid, Apprenticeship>>
+    {
+        public GetApprenticeships(Expression<Func<Apprenticeship, bool>> predicate = null)
```
gunndabad, line 10: We don't typically have constructors on DbQuery classes

src/Dfc.CourseDirectory.Functions/GetAllPublishedFrameworkApprenticeshipIds.cs
gunndabad: That's not the same as 'Published'

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/GetApprenticeships.cs
jameslawesynergy: Any particular reason? It seems like the nicest place to put this.

gunndabad: Consistency. Not super bothered; I foresee these types all becoming records soon :-)

src/Dfc.CourseDirectory.Functions/GetAllPublishedFrameworkApprenticeshipIds.cs
jameslawesynergy: Given the scope and intention of the story, I think we want to archive all Frameworks that aren't either `Archived` or `Deleted`. So I think the problem here is with the naming of the function, not the execution. What would you refer to for all Apprenticeships that are neither `Archived` or `Deleted`?

gunndabad: We don't really have a word for that. That said, we always exclude `Archived` and `Deleted` so for me that part is implied. I'd go with something like `GetApprenticeshipIdsWithFramework`.

gunndabad (approved):

src/Dfc.CourseDirectory.Functions/GetAllPublishedFrameworkApprenticeshipIds.cs
jameslawesynergy: I've renamed it to `GetAllFrameworkApprenticeshipIds` - Frameworks will be basically dead once this story is completed, as such so will this function.

PR #1721, opened 2020-11-03
Refactor FE and Apprenticeship authorization policies

gunndabad:


src/Dfc.CourseDirectory.Web/Startup.cs
```diff
@@ -161,14 +159,14 @@ public void ConfigureServices(IServiceCollection services)
...
-                    policy.RequireAssertion(x => (!x.User.IsInRole("Provider Superuser") && !x.User.IsInRole("Provider User")) ||
-                                                 x.User.Claims.Any(c => c.Type == "ProviderType" &&
-                                                                        _apprenticeshipClaims.Contains(c.Value))));
-                options.AddPolicy("Fe", policy =>
-                    policy.RequireAssertion(x => (!x.User.IsInRole("Provider Superuser") && !x.User.IsInRole("Provider User")) ||
-                                                                             x.User.Claims.Any(c => c.Type == "ProviderType" &&
-                                                                                                    _feClaims.Contains(c.Value, StringComparer.OrdinalIgnoreCase))));
+
+                options.AddPolicy(
+                    "Apprenticeship",
```
jameslawesynergy, line 164: Can we use a `const `or a `nameof()` here?

jameslawesynergy (approved):

soharadesign (approved):
🍻

src/Dfc.CourseDirectory.Web/Startup.cs
gunndabad: We have `AuthorizationPolicyNames` in V2 but I didn’t want to put it there because we shouldn’t be using this in V2 and I’m not super happy with these names (nor do I feel like trying to track down and fix the names here).

PR #1734, opened 2020-11-06
Port behaviours to filters

gunndabad:
Some time ago in a tech chat we spoke about moving cross-cutting things out of MediatR behaviors and into MVC filters. We now have a concrete reason to do so: we need the filters to be able to run on actions that don't use MediatR (e.g. actions that return only content).

This ports all except one of the current behaviors to filters. The final remaining one is a little more complicated and will be done separately.


src/Dfc.CourseDirectory.WebV2/Filters/AuthorizeAdminAttribute.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Dfc.CourseDirectory.WebV2.Security;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Dfc.CourseDirectory.WebV2.Filters
+{
+    public class AuthorizeAdminAttribute : ActionFilterAttribute
```
jameslawesynergy, line 7: Why not derivative of `IAuthorizationFIlter`?

src/Dfc.CourseDirectory.WebV2/Filters/AuthorizeCourseAttribute.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+
+            if (IsAuthorized())
+            {
+                var providerInfo = await providerInfoCache.GetProviderInfo(providerId.Value);
+                providerContextProvider.SetProviderContext(new ProviderContext(providerInfo));
+
+                await next();
+            }
+            else
```
jameslawesynergy, line 47: Invert condition to remove unnecessary else

src/Dfc.CourseDirectory.WebV2/Filters/RestrictProviderTypesAttribute.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Previous filter has already checked this is set
+            var providerContext = providerContextProvider.GetProviderContext();
+
+            if ((providerContext.ProviderInfo.ProviderType & PermittedProviderTypes) == 0)
+            {
+                // The active provider doesn't have any of the permitted provider types
+                throw new NotAuthorizedException();
+            }
+            else
```
jameslawesynergy, line 51: Unnecessary else

jameslawesynergy (approved):

src/Dfc.CourseDirectory.WebV2/Filters/AuthorizeAdminAttribute.cs
gunndabad: Relies on `NotAuthorizedException` being caught and handled by an exception filter. Exception filters don't catch errors from `IAuthorizationFilter`s, it seems.

src/Dfc.CourseDirectory.WebV2/Filters/AuthorizeCourseAttribute.cs
gunndabad: Done

src/Dfc.CourseDirectory.WebV2/Filters/RestrictProviderTypesAttribute.cs
gunndabad: Done

src/Dfc.CourseDirectory.WebV2/Filters/AuthorizeAdminAttribute.cs
jameslawesynergy: Why does it need to throw a `NotAuthorizedException`? Can it not just return the response from the filter?

gunndabad: It could but then we have the logic in two places instead of one.

PR #1749, opened 2020-11-11
PTCD-715 Fix Venues email address validation

jameslawesynergy:


gunndabad (commented):
Can we do this without calling an obsolete API (even if we copy/paste whatever Regex it's using internally)?

gunndabad (approved):

PR #1760, opened 2020-11-13
[PTCD-592] Cope gracefully with missing venue

timabell:
Cope gracefully with missing venue

We have some bad data that means venueIds are referenced that don't exist. The venue endpoint returns an empty body, which then becomes null, which then throws a null-ref exception which isn't very helpful.

By just not setting the value if it's null the page now loads and the user can correct the corrupted data.

Also corrected use of async.

# Before

Selection_20201113-01-83ebc83ebc83ebc

(hidden by generic error on other envs)

# After

Selection_20201113-01-8c4748c4748c474

click change...

Selection_20201113-01-87aaf87aaf87aaf

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
+++ b/src/Dfc.CourseDirectory.Web/Controllers/CourseSummaryController.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using Dfc.CourseDirectory.Services.CourseService;
 using Dfc.CourseDirectory.Services.Models.Courses;
 using Dfc.CourseDirectory.Services.Models.Regions;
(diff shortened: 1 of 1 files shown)


gunndabad (approved):

timabell:
Note that when testing against dev with ids from prod you can get a similar but unrelated error when the course doesn't exist in dev:

```
NullReferenceException: Object reference not set to an instance of an object.
Dfc.CourseDirectory.Web.Controllers.CourseSummaryController.Index(Nullable<Guid> courseId, Nullable<Guid> courseRunId) in CourseSummaryController.cs
...
courseRun = course.CourseRuns.Where(x => x.id == courseRunId.Value).FirstOrDefault();
```

soharadesign (approved):

timabell:
[Release 921 for testing](https://sfa-gov-uk.visualstudio.com/Digital%20First%20Careers/_releaseProgress?releaseId=17362&_a=release-pipeline-progress)

timabell:
Got thumbs up on slack from QA.
Rebased.

PR #1767, opened 2020-11-18
Debug ability to bypass QA

timabell:
To make it easier to work on the apprenticeship features during local development.

### DIFF
--- a/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderApprenticeshipQAStatusHandler.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderApprenticeshipQAStatusHandler.cs
@@ -20,7 +20,7 @@ WHERE ProviderId = @ProviderId";
             var paramz = new { query.ProviderId };
 
 #if DEBUG
-            if (bool.TryParse(System.Environment.GetEnvironmentVariable("BYPASS_QA"), out var bypassQA) && bypassQA)
+            if (bool.TryParse(System.Environment.GetEnvironmentVariable("CD_BYPASS_QA"), out var bypassQA) && bypassQA)
             {
                 return ApprenticeshipQAStatus.Passed;
             }
--- a/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderApprenticeshipQAStatusHandler.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderApprenticeshipQAStatusHandler.cs
@@ -20,7 +20,7 @@ WHERE ProviderId = @ProviderId";
             var paramz = new { query.ProviderId };
 
 #if DEBUG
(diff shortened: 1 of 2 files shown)


gunndabad (commented):
We tend to use 'CD_' prefix for environment variables.

Might be simpler to shim `ApprenticeshipQAStatusExtensions.ValueOrDefault` rather than this one query (to protect us if/when other queries also return this data).

src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderApprenticeshipQAStatusHandler.cs
```diff
@@ -19,6 +19,12 @@ public class GetProviderApprenticeshipQAStatusHandler :
 
             var paramz = new { query.ProviderId };
 
+#if DEBUG
+            if (System.Environment.GetEnvironmentVariable("BYPASS_QA")!=null && bool.Parse((System.Environment.GetEnvironmentVariable("BYPASS_QA"))))
```
gunndabad, line 23:
```suggestion
            if (bool.TryParse(System.Environment.GetEnvironmentVariable("BYPASS_QA", out var bypassQA) && bypassQA)
```

timabell: Cheers. Rebased and switched to new out-var syntax.

timabell:
> We tend to use 'CD_' prefix for environment variables.

Sure; is there a reason for namespacing them that I'm unaware of?

> Might be simpler to shim ApprenticeshipQAStatusExtensions.ValueOrDefault rather than this one query (to protect us if/when other queries also return this data).

I tried modifying [ValueOrDefault](https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/a87afc044d9987fb57348b78ee191c44eae4c6dc/src/Dfc.CourseDirectory.Core/Models/ApprenticeshipQAStatus.cs#L42), I got the "passed" notification on the dashboard but not the links to the apprenticeships so I guess not everything is using that extension method currently.

gunndabad:
> > We tend to use 'CD_' prefix for environment variables.
> 
> Sure; is there a reason for namespacing them that I'm unaware of?
> 
The .NET ones are prefixed so I followed the same convention. I assume it's so their purpose is obviously linked to a particular application but not sure.
> > Might be simpler to shim ApprenticeshipQAStatusExtensions.ValueOrDefault rather than this one query (to protect us if/when other queries also return this data).
> 
> I tried modifying [ValueOrDefault](https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/a87afc044d9987fb57348b78ee191c44eae4c6dc/src/Dfc.CourseDirectory.Core/Models/ApprenticeshipQAStatus.cs#L42), I got the "passed" notification on the dashboard but not the links to the apprenticeships so I guess not everything is using that extension method currently.
> 
Fair enough.
> Selection_20201123-01-c639cc639cc639c

gunndabad (approved):

PR #1770, opened 2020-11-18
Remove ProviderService

jameslawesynergy:
Removing `ProviderService` proxy from `Services` project that calls the external UKRLP API for data operations.


src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
```diff
@@ -0,0 +1,13 @@
...
+using OneOf;
+using OneOf.Types;
+
+namespace Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries
+{
+    public class UpdateProviderOnboarded : ICosmosDbQuery<OneOf<NotFound, Success>>
+    {
+        public Guid ProviderId { get; set; }
+
+        public string UpdatedBy { get; set; }
```
gunndabad, line 11:
```suggestion
        public UserInfo UpdatedBy { get; set; }
```

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/UpdateProviderBulkUploadStatusHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+            try
+            {
+                var response = await client.ReadDocumentAsync<Provider>(documentUri);
+
+                var provider = response.Document;
+
+                if (provider.BulkUploadStatus == null)
+                {
+                    provider.BulkUploadStatus = new ProviderBulkUploadStatus();
+                }
```
gunndabad, line 30:
```suggestion
                provider.BulkUploadStatus ??= new ProviderBulkUploadStatus();
```

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/UpdateProviderBulkUploadStatusHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+
+                provider.BulkUploadStatus.PublishInProgress = request.PublishInProgress;
```
gunndabad, line 32: Do we want to set the `Updated`* fields here too?

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
```diff
@@ -0,0 +1,13 @@
...
+    }
```
gunndabad, line 12: Add `DateTime UpdatedDate` property on here (otherwise it's harder to test with our `IClock` stuff)

src/Dfc.CourseDirectory.Web/Controllers/BulkUploadApprenticeshipsController.cs
```diff
@@ -25,64 +25,29 @@ namespace Dfc.CourseDirectory.Web.Controllers
     [RestrictApprenticeshipQAStatus(ApprenticeshipQAStatus.Passed)]
     public class BulkUploadApprenticeshipsController : Controller
     {
+        private const string _blobContainerPath = "/Apprenticeship Bulk Upload/Files/";
```
gunndabad, line 28:
```suggestion
        private const string BlobContainerPath = "/Apprenticeship Bulk Upload/Files/";
```

gunndabad:
Are there any config parameters we can remove from the ARM templates?

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
jameslawesynergy: What about injecting `IClock`? (Isn't there a .net abstraction of this now, `ISystemClock`?)

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/UpdateProviderBulkUploadStatusHandler.cs
jameslawesynergy: I' prefer to leave them for the time being and implement them when we need them in CD. (Trying to stick to like-for-like as much as possible to reduce risk.)

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
jameslawesynergy: Doing so will change the behaviour as we currently pass through a magic string - was your intention here to pass the full `UserInfo` object or future proof?

gunndabad: It's for consistency and future-proofing. Every other handler takes a `UserInfo` property on the command for setting the `{Created|Updated}By` field.

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/UpdateProviderBulkUploadStatusHandler.cs
gunndabad: Fair enough

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
gunndabad: As above, current handlers that do this have a property for the `{Created|Updated}By` and that's populated by the caller (by referencing an `IClock`) e.g. https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/main/src/Dfc.CourseDirectory.WebV2/Features/NewApprenticeshipProvider/ApprenticeshipSummary.cs#L163

gunndabad: There's more than one `ISystemClock` and none of them are in the BCL

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
jameslawesynergy: Implementing that properly will change the behaviour. Do we want to do this?

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
jameslawesynergy: I meant injecting `IClock` directly into the handler? What's the advantages over passing it as a param?

gunndabad: Not so much that it's any better or worse, more that it's consistent with everywhere else.

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
gunndabad: For me yes. `UpdatedBy` should be the email of the user making the change.

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/UpdateProviderOnboarded.cs
jameslawesynergy: Updated

gunndabad (approved):

PR #1779, opened 2020-11-23
Add a LastSyncedFromCosmos column to SQL tables

gunndabad:
We have some invalid data in our SQL tables currently due to some hard deletes on Cosmos data (the hard deletes have since been fixed). By adding this column so we can identify those records that no longer exist in Cosmos after we re-run the 'sync all' processes and seeing which records haven't been updated.

It also may be useful for the forthcoming Azure Search indexer as the 'high water mark' column (https://docs.microsoft.com/en-us/azure/search/search-howto-connecting-azure-sql-database-to-azure-search-using-indexers#high-water-mark-change-detection-policy)


timabell (approved):
:ship:

PR #1784, opened 2020-11-25
PTCD-756 Add a temporary cleanup function for removing zombie courses from SQL

gunndabad:
These were added to Cosmos then hard deleted so our sync process didn't see' their removal. This uses the new LastSyncedFromCosmos column for figuring out what to delete (which will always be null for these zombie courses). A full sync 'SyncCoursesCosmosCollectionToSql' should be run before running this to ensure legit entries have the LastSyncedFromCosmos column populated.

### DIFF
--- a/src/Dfc.CourseDirectory.Core/SqlDataSync.cs
+++ b/src/Dfc.CourseDirectory.Core/SqlDataSync.cs
@@ -17,7 +17,7 @@ namespace Dfc.CourseDirectory.Core
 {
     public class SqlDataSync
     {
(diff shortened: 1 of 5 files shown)


jameslawesynergy:
Is it worth forcing a full sync as part of this call if it's an absolute requirement?

jameslawesynergy (approved):

gunndabad:
> Is it worth forcing a full sync as part of this call if it's an absolute requirement?

The full-sync jobs have been a little flaky in the past so I thought keeping them separate might make it easier to deploy.

PR #1790, opened 2020-11-27
Add ADR stub

timabell:
We're about to make a decision on sql deploys, so this would be a good place to record it.


timabell:
The decision has been punted for now so closing this PR

PR #1801, opened 2020-12-01
PTCD-763 Add Azure Function for removing duplicate provider records from Cosmos

gunndabad:
Also included a read-only 'dry run' function.

### DIFF
--- a/src/Dfc.CourseDirectory.Functions/DeleteDuplicateProviderRecords.cs
+++ b/src/Dfc.CourseDirectory.Functions/DeleteDuplicateProviderRecords.cs
@@ -115,6 +115,22 @@ where x.ProviderId is null and y.ProviderType <> p.ProviderType
 
 if @DryRun = 0
 begin
+    -- Reassign User->Provider mappings for provider IDs that are being removed
+	;with UserProvidersCte as (
+		select up.UserId, up.ProviderId OldProviderId, z.ProviderId NewProviderId
+		from Pttcd.UserProviders up
+		join #DeletingProviderIds x ON up.ProviderId = x.ProviderId
+		join Pttcd.Providers z ON x.Ukprn = z.Ukprn
+		left join #DeletingProviderIds y on z.ProviderId = y.ProviderId
+		where y.ProviderId is null
+	)
+	merge Pttcd.UserProviders as target
+	using (select * from UserProvidersCte) as source
+	on target.UserId = source.UserId and target.ProviderId = source.OldProviderId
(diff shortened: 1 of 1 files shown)


src/Dfc.CourseDirectory.Functions/DeleteDuplicateProviderRecords.cs
```diff
@@ -0,0 +1,198 @@
...
+            IList<UpdatedProviderInfo> updatedProviders;
+            IList<DeletedProviderInfo> deletedProviders;
+
+            using (var reader = await sqlConnection.QueryMultipleAsync(sql, new { DryRun = dryRun }))
+            {
+                updatedProviders = (await reader.ReadAsync<UpdatedProviderInfo>()).AsList();
+                deletedProviders = (await reader.ReadAsync<DeletedProviderInfo>()).AsList();
+            }
+
+            if (!dryRun)
```
timabell, line 155: Personally I'd invert the `if` and early-`return` here to reduce nesting even though it means duplicating the Ok somewhat

src/Dfc.CourseDirectory.Functions/DeleteDuplicateProviderRecords.cs
```diff
@@ -0,0 +1,198 @@
...
+                foreach (var record in updatedProviders)
+                {
+                    await _cosmosDbQueryDispatcher.ExecuteQuery(new UpdateProviderType()
+                    {
+                        ProviderId = record.ProviderId,
+                        ProviderType = record.NewProviderType
+                    });
+                }
+
+                foreach (var record in deletedProviders)
```
timabell, line 166: what happens if the sql fails having deleted some, would we then have to work harder to fixup cosmos?

timabell (approved):

soharadesign (approved):

timabell:
might be worth wrapping the sql in
```
set xact_abort on
begin tran
...
commit
```

src/Dfc.CourseDirectory.Functions/DeleteDuplicateProviderRecords.cs
gunndabad: We could re-run the full sync to get records back into SQL then re-run. Given that we don't have distributed transactions it's never going to be perfect.

gunndabad:
> might be worth wrapping the sql in
> 
> ```
> set xact_abort on
> begin tran
> ...
> commit
> ```

Ah yeh I missed re-adding the transaction.

PR #1802, opened 2020-12-01
PTCD-743 Grant Provider T Level Status

jameslawesynergy:
Added ProviderType.TLevels and updated EditProviderType to allow selection of TLevels ProviderType
Added TLevelDefinitions and ProviderTLevelDefinitions to SQL database
Added configure TLevelDefinitions for Provider functionality

### DIFF
--- a/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/UpsertProviderTLevelDefinitionsHandler.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/UpsertProviderTLevelDefinitionsHandler.cs
@@ -40,7 +40,7 @@ WHEN NOT MATCHED THEN
         @ProviderId,
         source.TLevelDefinitionId
     )
-WHEN NOT MATCHED BY SOURCE THEN
+WHEN NOT MATCHED BY SOURCE AND target.ProviderId = @ProviderId THEN
     DELETE;";
 
             await transaction.Connection.ExecuteAsync(
--- a/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/UpsertProviderTLevelDefinitionsHandler.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/UpsertProviderTLevelDefinitionsHandler.cs
@@ -1,0 +1,54 @@
+﻿using System.Data.SqlClient;
+using System.Linq;
+using System.Threading.Tasks;
+using Dapper;
+using Dfc.CourseDirectory.Core.DataStore.Sql.Queries;
+using OneOf.Types;
+
+namespace Dfc.CourseDirectory.Core.DataStore.Sql.QueryHandlers
+{
+    public class UpsertProviderTLevelDefinitionsHandler : ISqlQueryHandler<UpsertProviderTLevelDefinitions, None>
+    {
+        public async Task<None> Execute(SqlTransaction transaction, UpsertProviderTLevelDefinitions query)
+        {
+            const string createTableSql = @"
+CREATE TABLE #TLevelDefinitionIds (
+    TLevelDefinitionId UNIQUEIDENTIFIER
+)";
+
+            await transaction.Connection.ExecuteAsync(createTableSql, transaction: transaction);
+
+            await BulkCopyHelper.WriteRecords(
+                query.TLevelDefinitionIds.Distinct().Select(id => new { TLevelDefinitionId = id }),
+                tableName: "#TLevelDefinitionIds",
+                transaction);
+
+            const string sql = @"
+MERGE Pttcd.ProviderTLevelDefinitions AS target
+USING (
+    SELECT
+        TLevelDefinitionId
+    FROM #TLevelDefinitionIds
+) AS source
+ON target.ProviderId = @ProviderId
+AND target.TLevelDefinitionId = source.TLevelDefinitionId
+WHEN NOT MATCHED THEN
+    INSERT (
+        ProviderId,
+        TLevelDefinitionId
+    ) VALUES (
+        @ProviderId,
+        source.TLevelDefinitionId
+    )
+WHEN NOT MATCHED BY SOURCE THEN
+    DELETE;";
+
+            await transaction.Connection.ExecuteAsync(
+                sql,
+                param: new { query.ProviderId },
+                transaction: transaction);
+
+            return new None();
+        }
+    }
+}
--- a/src/Dfc.CourseDirectory.Database/Dfc.CourseDirectory.Database.sqlproj
+++ b/src/Dfc.CourseDirectory.Database/Dfc.CourseDirectory.Database.sqlproj
@@ -242,11 +242,13 @@
     <Build Include="Pttcd\Indexes\IX_Courses_ProviderUkprn.sql" />
     <Build Include="Pttcd\Tables\Regions.sql" />
     <Build Include="Pttcd\Tables\CourseRunSubRegions.sql" />
+    <Build Include="Pttcd\Tables\ProviderTLevelDefinitions.sql" />
+    <Build Include="Pttcd\Tables\TLevelDefinitions.sql" />
   </ItemGroup>
   <ItemGroup>
     <RefactorLog Include="dfc-coursedirectory.refactorlog" />
   </ItemGroup>
   <ItemGroup>
-    <PostDeploy Include="Scripts\Script.Regions.sql" />
+    <PostDeploy Include="Scripts\Script.PostDeploy.sql" />
   </ItemGroup>
 </Project>
--- a/src/Dfc.CourseDirectory.Database/Pttcd/Tables/ProviderTLevelDefinitions.sql
+++ b/src/Dfc.CourseDirectory.Database/Pttcd/Tables/ProviderTLevelDefinitions.sql
@@ -1,0 +1,5 @@
+﻿CREATE TABLE [Pttcd].[ProviderTLevelDefinitions] (
+    [ProviderId]          UNIQUEIDENTIFIER NOT NULL,
+    [TLevelDefinitionId]  UNIQUEIDENTIFIER NOT NULL
+);
+
--- a/src/Dfc.CourseDirectory.Database/Pttcd/Tables/ProviderTLevelDefinitions.sql
+++ b/src/Dfc.CourseDirectory.Database/Pttcd/Tables/ProviderTLevelDefinitions.sql
@@ -1,5 +1,8 @@
 ﻿CREATE TABLE [Pttcd].[ProviderTLevelDefinitions] (
-    [ProviderId]          UNIQUEIDENTIFIER NOT NULL,
-    [TLevelDefinitionId]  UNIQUEIDENTIFIER NOT NULL
+    [ProviderId]         UNIQUEIDENTIFIER NOT NULL,
+    [TLevelDefinitionId] UNIQUEIDENTIFIER NOT NULL,
+    CONSTRAINT [PK_ProviderTLevelDefinitions] PRIMARY KEY CLUSTERED ([ProviderId] ASC, [TLevelDefinitionId] ASC)
 );
+
+
 
--- a/src/Dfc.CourseDirectory.Database/Pttcd/Tables/TLevelDefinitions.sql
+++ b/src/Dfc.CourseDirectory.Database/Pttcd/Tables/TLevelDefinitions.sql
@@ -1,0 +1,8 @@
+﻿CREATE TABLE [Pttcd].[TLevelDefinitions] (
+    [TLevelDefinitionId] UNIQUEIDENTIFIER NOT NULL,
+    [FrameworkCode]      INT              NOT NULL,
+    [ProgType]           INT              NOT NULL,
+    [Name]               NVARCHAR (MAX)   NOT NULL,
+    CONSTRAINT [PK_Pttcd.TLevelDefinitions] PRIMARY KEY CLUSTERED ([TLevelDefinitionId] ASC)
+);
+
--- a/src/Dfc.CourseDirectory.Database/Scripts/Script.PostDeploy.sql
+++ b/src/Dfc.CourseDirectory.Database/Scripts/Script.PostDeploy.sql
@@ -162,4 +162,48 @@ BEGIN
 	('E10000009', 'Dorset', 'E12000009'),
 	('E10000013', 'Gloucestershire', 'E12000009'),
 	('E10000027', 'Somerset', 'E12000009')
-END
+END;
+
+DECLARE @TLevelDefinitions AS table (
+	[TLevelDefinitionId] [uniqueidentifier] NOT NULL,
+	[FrameworkCode] [int] NOT NULL,
+	[ProgType] [int] NOT NULL,
+	[Name] [nvarchar](max) NOT NULL
+)
+
+INSERT INTO @TLevelDefinitions (
+	[TLevelDefinitionId],
+	[FrameworkCode],
+	[ProgType],
+	[Name]
+)
+VALUES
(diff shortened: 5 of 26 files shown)


src/Dfc.CourseDirectory.Database/Dfc.CourseDirectory.Database.sqlproj
```diff
@@ -240,11 +241,13 @@
...
     <Build Include="Pttcd\Tables\Regions.sql" />
+    <Build Include="Pttcd\Tables\ProviderTLevelDefinitions.sql" />
+    <Build Include="Pttcd\Tables\TLevelDefinitions.sql" />
   </ItemGroup>
   <ItemGroup>
     <RefactorLog Include="dfc-coursedirectory.refactorlog" />
   </ItemGroup>
   <ItemGroup>
-    <PostDeploy Include="Scripts\Script.Regions.sql" />
+    <PostDeploy Include="Scripts\Script.PostDeploy.sql" />
```
jameslawesynergy, line 251: Apparently you can only have one PostDeploy script.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/ProviderDetailsTests.cs
```diff
@@ -90,6 +92,45 @@ public async Task Get_ValidRequest_RendersExpectedContent(TestUserType userType)
...
+            // Act
+            var response = await HttpClient.SendAsync(request);
+
+            // Assert
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+
+            var doc = await response.GetDocument();
+            var providerTypeValue = doc.GetSummaryListValueWithKey("Provider type");
+
+            providerTypeValue.Should().Contain("T Levels");
```
jameslawesynergy, line 121: If you know a better way to assert this I'd love to hear it.

src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/UpsertProviderTLevelDefinitionsHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+AND target.TLevelDefinitionId = source.TLevelDefinitionId
+WHEN NOT MATCHED THEN
+    INSERT (
+        ProviderId,
+        TLevelDefinitionId
+    ) VALUES (
+        @ProviderId,
+        source.TLevelDefinitionId
+    )
+WHEN NOT MATCHED BY SOURCE THEN
```
gunndabad, line 43:
```suggestion
WHEN NOT MATCHED BY SOURCE AND target.ProviderId = @ProviderId THEN
```

src/Dfc.CourseDirectory.Database/Pttcd/Tables/ProviderTLevelDefinitions.sql
```diff
@@ -0,0 +1,5 @@
+﻿CREATE TABLE [Pttcd].[ProviderTLevelDefinitions] (
+    [ProviderId]          UNIQUEIDENTIFIER NOT NULL,
```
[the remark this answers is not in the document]
gunndabad: > Pondering composite PK here...?

Agreed

src/Dfc.CourseDirectory.Database/Scripts/Script.PostDeploy.sql
```diff
@@ -162,4 +162,53 @@ BEGIN
 	('E10000009', 'Dorset', 'E12000009'),
 	('E10000013', 'Gloucestershire', 'E12000009'),
 	('E10000027', 'Somerset', 'E12000009')
-END
+END;
+
+IF  EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[Pttcd].[TLevelDefinitions]') AND type in (N'U'))
```
gunndabad, line 167: Not sure this check is required

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
```diff
@@ -40,25 +52,77 @@ public async Task<Command> Handle(Query request, CancellationToken cancellationT
...
+                    When(c => c.SelectedProviderTLevelDefinitionIds?.Any() ?? false, () =>
+                    {
+                        RuleFor(c => c.SelectedProviderTLevelDefinitionIds)
+                            .Must(ids => ids.All(id => tLevelDefinitions.Any(d => d.TLevelDefinitionId == id)));
+                    });
+                });
+            }
+        }
+
+        private static IEnumerable<ProviderTLevelDefinitionViewModel> ProviderTLevelDefinitions(
```
gunndabad, line 119:
```suggestion
        private static IEnumerable<ProviderTLevelDefinitionViewModel> GetProviderTLevelDefinitions(
```

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
```diff
@@ -14,22 +22,26 @@ public class Query : IRequest<Command>
         public Guid ProviderId { get; set; }
     }
 
-    public class Command : IRequest<Success>
+    public class Command : IRequest<OneOf<ModelWithErrors<Command>, Success>>
```
gunndabad, line 25: I think the read and write model should be split up here - `ProviderTLevelDefinitions` shouldn't be in the write model, for me

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
```diff
@@ -40,25 +52,77 @@ public async Task<Command> Handle(Query request, CancellationToken cancellationT
...
+                return new ModelWithErrors<Command>(request, validationResult);
+            }
+
             await _cosmosDbQueryDispatcher.ExecuteQuery(new UpdateProviderType()
             {
                 ProviderId = request.ProviderId,
                 ProviderType = request.ProviderType
             });
 
+            await _sqlQueryDispatcher.ExecuteQuery(new SqlQueries.UpsertProviderTLevelDefinitions
```
gunndabad, line 86: Isn't there a requirement to maintain the T-Level Definitions for a provider even when T-Levels is removed?

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
```diff
@@ -40,25 +52,77 @@ public async Task<Command> Handle(Query request, CancellationToken cancellationT
...
+        {
+            public CommandValidator(IEnumerable<SqlModels.TLevelDefinition> tLevelDefinitions)
+            {
+                When(c => c.ProviderType.HasFlag(ProviderType.TLevels), () =>
+                {
+                    RuleFor(c => c.SelectedProviderTLevelDefinitionIds)
+                        .NotEmpty()
+                        .WithMessage("Select the T Levels this provider can offer");
+
+                    When(c => c.SelectedProviderTLevelDefinitionIds?.Any() ?? false, () =>
```
gunndabad, line 110: I think this needs a message if we're going to add this rule.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/EditProviderTypeTests.cs
```diff
@@ -77,14 +80,21 @@ public async Task Get_ValidRequestNoneProviderType_RendersExpectedOutput()
...
+        [InlineData(ProviderType.Apprenticeships | ProviderType.FE | ProviderType.TLevels, new[] { "fe", "apprenticeships", "tLevels" })]
+
         public async Task Get_ValidRequest_RendersExpectedOutput(
             ProviderType providerType,
             IEnumerable<string> expectedCheckedTestIds)
         {
             // Arrange
             var providerId = await TestData.CreateProvider(providerType: providerType);
 
+            var tLevelDefinitionIds = await Task.WhenAll(Enumerable.Range(0, 3).Select(_ => TestData.CreateTLevelDefinition()));
```
gunndabad, line 96: I think `frameworkCode` & `progType` pairs are meant to be unique. If that's the case I think we should reflect that in the test data here.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/EditProviderTypeTests.cs
```diff
@@ -192,5 +254,163 @@ public async Task Post_ValidRequest_UpdatesProviderTypeAndRedirects(ProviderType
...
+
+            var doc = await response.GetDocument();
+            doc.AssertHasError(nameof(Command.SelectedProviderTLevelDefinitionIds), "Select the T Levels this provider can offer");
+
+            CosmosDbQueryDispatcher.VerifyExecuteQuery<UpdateProviderType, OneOf<NotFound, Success>>(q =>
+                q.ProviderId == providerId && q.ProviderType == providerType, Times.Never());
+        }
+
+        [Fact]
+        public async Task Post_WithTLevelsProviderAndInvalidTLevelDefinitionId_()
```
gunndabad, line 335: Nit:
```suggestion
        public async Task Post_WithTLevelsProviderAndInvalidTLevelDefinitionId()
```

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/ProviderDetailsTests.cs
gunndabad: Perhaps a similar method to `GetSummaryListValueWithKey` that returns the `IElement` instead of its `TextContent` then could loop over the `<li>`s

gunndabad: `GetSummaryListValueElementWithKey`?

src/Dfc.CourseDirectory.Database/Pttcd/Tables/TLevelDefinitions.sql
```diff
@@ -0,0 +1,8 @@
+﻿CREATE TABLE [Pttcd].[TLevelDefinitions] (
+    [TLevelDefinitionId] UNIQUEIDENTIFIER NOT NULL,
+    [FrameworkCode]      INT              NOT NULL,
```
gunndabad, line 3: If `FrameworkCode` and `ProgType` pairs unique a constraint to enforce it would be good here

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
gunndabad: Alternatively, could remove any invalid entries in `SelectedProviderTLevelDefinitionIds` early on to ignore them.

src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/UpsertProviderTLevelDefinitionsHandler.cs
jameslawesynergy: Any particular reason for this? This condition is given in the main using statement and removing it from there would affect `WHEN NOT MATCHED`.

gunndabad: Pretty sure it will remove any entries for other providers as written

jameslawesynergy: That's surprising, good spot

src/Dfc.CourseDirectory.Database/Pttcd/Tables/TLevelDefinitions.sql
jameslawesynergy: I don't see the value of doing that at this time, especially as we're currently controlling data ingress via script currently.

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
jameslawesynergy: I was under the impression that it was only possible to bind a single model to an MVC view - happy to go through this offline.

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
jameslawesynergy: Not in this story! (I've played with it and given the validation requirements this is the only behaviour that makes sense.)

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
jameslawesynergy: This is a data integrity validation - the message is the default (see unit tests), is adequately descriptive and will not be shown to users under typical failure scenarios. (They'd have to fudge the request or have a data race condition which is unlikely)

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
jameslawesynergy: Any particular reason?

gunndabad: Method names should be commands, IMO. This reads like a property.

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
gunndabad: In other places where we do similar data integrity checks we just throw `InvalidStateException` and let the default error page show. No reason to display a pretty message in these situations, I think. All our current validation messages are curated; this blurs that line.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/EditProviderTypeTests.cs
jameslawesynergy: I'm not sure we know enough about the data at this stage to bother with this - we can look at this again once we're actually using that particular data.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/EditProviderTypeTests.cs
jameslawesynergy: Typo! :)

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/ProviderDetailsTests.cs
jameslawesynergy: That perhaps would work but it were not for the way in which we're displaying the list of TLevelDefinitions (using a govuk-details component) - going down this route would make the test considerably more fragile and too closely tied to the implementation IMO.

src/Dfc.CourseDirectory.WebV2/Features/Providers/EditProviderType.cs
jameslawesynergy: I'd rather keep all of the model validation logic in a single place, however I'm not averse to it throwing an exception. Although I would argue that is a sub-optimal behaviour from the user's point of view given that there may be a change in the underlying data between view render and post that causes the error but is more recoverable with an error message rather than a generic exception page.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Providers/ProviderDetailsTests.cs
gunndabad: Could put a `data-testid` attribute on the component and just check if it exists?

gunndabad (approved):

PR #1803, opened 2020-12-02
Upgrade to latest FormFlow package & react to API changes

gunndabad:
Mostly renames - '{Form}Flow` is now called 'Journey'.

Brings in functionality to use both a route parameter and a unique GUID to identify a journey instance. This is required for the Add T-Level journey as well as for fixing some QA and Venue-related bugs indentified recently.


timabell:
We really do need to get the author of this package to document it so it's maintainable for more than just themselves. This is currently a bit of a risk for the project because if the person who knows how to use it goes we will be a bit stuck. https://github.com/gunndabad/formflow - I hear we know them well :grin: so maybe we have some leverage :thinking:

gunndabad:
> We really do need to get the author of this package to document it so it's maintainable for more than just themselves. This is currently a bit of a risk for the project because if the person who knows how to use it goes we will be a bit stuck. https://github.com/gunndabad/formflow - I hear we know them well :grin: so maybe we have some leverage :thinking: 

It's next after the tag helper library 😁

timabell (approved):
Approving to keep moving, but noting above concerns.

PR #1821, opened 2020-12-08
Add helper method for creating a RouteValues dictionary from the current request's query params

gunndabad:


src/Dfc.CourseDirectory.WebV2/ServiceCollectionExtensions.cs
```diff
@@ -160,6 +160,7 @@ public static class ServiceCollectionExtensions
             services.Configure<GoogleAnalyticsOptions>(configuration.GetSection("GoogleAnalytics"));
             services.Configure<GoogleTagManagerOptions>(configuration.GetSection("GoogleTagManager"));
             services.AddTransient<SqlDataSync>();
+            services.AddTransient<RouteValuesHelper>();
```
jameslawesynergy, line 163: Surely this would be scoped, requiring scope for the `HttpContext`?

src/Dfc.CourseDirectory.WebV2/ViewHelpers/RouteValuesHelper.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+
+namespace Dfc.CourseDirectory.WebV2.ViewHelpers
+{
+    public class RouteValuesHelper
```
jameslawesynergy, line 7: I loathe "Helpers" 😠

jameslawesynergy (approved):

src/Dfc.CourseDirectory.WebV2/ViewHelpers/RouteValuesHelper.cs
gunndabad: Open to better names :-)

src/Dfc.CourseDirectory.WebV2/ServiceCollectionExtensions.cs
gunndabad: Don't think it will make a difference here but I've changed it

PR #1824, opened 2020-12-08
Add authorizedTLevelDefinitionIds parameter to TestData.CreateProvider

gunndabad:


jameslawesynergy (changes requested):
I think we already have this, albeit with a different terminology - I called the `ProviderTLevelDefinitions` - I'd rather stick with that naming convention than confuse the codebase, especially with `Authorize` being a keyword within the framework.

gunndabad:
> Unsure why you've renamed `ProviderTLevelDefinitions` - I'd rather stick with that naming convention that's consistent throughout layers than confuse the codebase, especially with `Authorize` being a keyword within the framework.

`ProviderTLevelDefinitions` doesn't mean much to me. I'm open to better names than `AuthorizedTLevelDefinitions`, though.

> Also, why closely link `ProviderTLevelDefinitions` with `Provider`? I believe we've already got the test logic set up to allow these to be set independently?

It makes setting up a T Level provider considerably less tedious (and we have no method on `TestData` at all today for doing it).

jameslawesynergy:
> `ProviderTLevelDefinitions` doesn't mean much to me. I'm open to better names than `AuthorizedTLevelDefinitions`, though.

It means a lot to me: "Provider" is a domain entity, "TLevelDefinition" is a domain entity. Combine them together and that tells me that it's a lookup with a many-to-many relationship, it also tells me where I can find them in the database 😄 We can have a chat tomorrow offline or discuss with the team during tech chat if you like?

> (and we have no method on `TestData` at all today for doing it).
https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/4223c3c01de2c0a231cc269bfd2c06e98c4d3dff/tests/Dfc.CourseDirectory.Testing/TestData/TestData.CreateTLevelDefinition.cs#L9

PR #1833, opened 2020-12-14
Use enum values in Testing DataStore GetApprenticeshipsByIdsHandler

timabell:
Use enum values in Testing DataStore GetApprenticeshipsByIdsHandler

Just a suggestions, I might have missed a reason why we're not doing this.

### DIFF
--- a/tests/Dfc.CourseDirectory.Testing/DataStore/CosmosDb/QueryHandlers/GetApprenticeshipsByIdsHandler.cs
+++ b/tests/Dfc.CourseDirectory.Testing/DataStore/CosmosDb/QueryHandlers/GetApprenticeshipsByIdsHandler.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Models;
 using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries;
+using Dfc.CourseDirectory.Core.Models;
 
 namespace Dfc.CourseDirectory.Testing.DataStore.CosmosDb.QueryHandlers
 {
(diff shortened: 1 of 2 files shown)


tests/Dfc.CourseDirectory.Testing/DataStore/CosmosDb/QueryHandlers/GetApprenticeshipsByIdsHandler.cs
```diff
@@ -12,7 +13,7 @@ public class GetApprenticeshipsByIdsHandler
         public IDictionary<Guid, Apprenticeship> Execute(InMemoryDocumentStore inMemoryDocumentStore, GetApprenticeshipsByIds request) =>
             inMemoryDocumentStore.Apprenticeships.All
                 .Where(p => request.ApprenticeshipIds.Contains(p.Id))
-                .Where(p => p.RecordStatus != 4 && p.RecordStatus != 8)
+                .Where(p => p.RecordStatus != (int)RecordStatus.Archived && p.RecordStatus != (int)RecordStatus.Deleted)
```
gunndabad, line 16: Better to use `ApprenticeshipStatus` here; `RecordStatus` is from the old models and shouldn't be used going forward

tests/Dfc.CourseDirectory.Testing/Dfc.CourseDirectory.Testing.csproj
```diff
@@ -20,6 +20,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\Dfc.CourseDirectory.Core\Dfc.CourseDirectory.Core.csproj" />
+    <ProjectReference Include="..\..\src\Dfc.CourseDirectory.Services\Dfc.CourseDirectory.Services.csproj" />
```
gunndabad, line 23: Don't really want old stuff referenced from new stuff if we can avoid it.

tests/Dfc.CourseDirectory.Testing/DataStore/CosmosDb/QueryHandlers/GetApprenticeshipsByIdsHandler.cs
timabell: Ah, maybe we should use Obsolete to signpost to the new one, or at least a comment on the old one. Thanks

gunndabad: `[Obsolete]` will fail build currently

timabell: Yeah, changed my mind immediately, will also be very noise if we obsolete everything in legacy.

timabell:
rebase + squash pushed....

PR #1834, opened 2020-12-14
Improve debuggability of integration tests

gunndabad:
By default TestServer allows unhandled exceptions to bubble up to the Request sender (HttpClient.SendAsync in our case). This is useful for debugging test failures since you get the stack trace from the source exception in failure messages instead of the stack trace from HttpResponseMessage.EnsureSuccessCode (or similar). Our custom error handling is masking this, however.

Given that we don't rely on our custom error behaviours in our tests (indeed we disable one of its behaviours) turn off custom errors for tests.

### DIFF
--- a/src/Dfc.CourseDirectory.WebV2/Features/Error/ErrorController.cs
+++ b/src/Dfc.CourseDirectory.WebV2/Features/Error/ErrorController.cs
@@ -8,7 +8,7 @@ namespace Dfc.CourseDirectory.WebV2.Features.Error
     [Route("error")]
     public class ErrorController : Controller
     {
-        public IActionResult Error(int? code, [FromServices] Settings settings)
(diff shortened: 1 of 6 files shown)


timabell (approved):
It seems your commit does a bit more than your PR description would imply, what's the change of `RewriteForbiddenToNotFound` for? Perhaps split to two commits if it's just cleanup?

gunndabad:
> It seems your commit does a bit more than your PR description would imply, what's the change of `RewriteForbiddenToNotFound` for? Perhaps split to two commits if it's just cleanup?

`RewriteForbiddenToNotFound` used to be required to, effectively, disable the custom error handling bits (or at least disable their noticeable effects). Given the change to remove that error handling, the config is no longer required.

PR #1856, opened 2021-01-05
Fix SetProviderDisplayNameSourceHandler return value

gunndabad:
Now that FindACourseIndex table is updated via a trigger the row count from a single row update is no longer necessarily 1. Invert the condition in this handler to account for this.


src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/SetProviderDisplayNameSourceHandler.cs
```diff
@@ -20,7 +20,7 @@ public class SetProviderDisplayNameSourceHandler :
                 ProviderId = query.ProviderId
             };
 
-            var updated = (await transaction.Connection.ExecuteAsync(sql, paramz, transaction)) == 1;
+            var updated = (await transaction.Connection.ExecuteAsync(sql, paramz, transaction)) != 0;
```
jameslawesynergy, line 23: `> 0`?

jameslawesynergy (approved):

PR #1869, opened 2021-01-14
Filter out archived venues in GetVenuesByProviderHandler

timabell:
As discussed on slack, there is no use case for returning archived Venues

### DIFF
--- a/src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/GetVenuesByProviderHandler.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/GetVenuesByProviderHandler.cs
@@ -3,6 +3,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Models;
 using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries;
+using Dfc.CourseDirectory.Core.Models;
 using Microsoft.Azure.Documents.Client;
 using Microsoft.Azure.Documents.Linq;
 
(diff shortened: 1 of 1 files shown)


jameslawesynergy:
> As discussed on slack, there is no use case for returning archived Venues

Have you been through the code and looked at all of the consumers of the Handler and sense-checked for returning only non-archived `Venues`?



Also, weren't there others that needed fixing up in the same way?

timabell:
> Have you been through the code and looked at all of the consumers of the Handler and sense-checked for returning only non-archived `Venues`?

I got the impression @gunndabad already knew that it would be okay, but perhaps I should. Good call.

> Also, weren't there others that needed fixing up in the same way?

Possibly, but I'm inclined to get this out and then worry about the others if needed.

jameslawesynergy:
I was under the impression from the slack conversation [https://dfc-beta.slack.com/archives/G01CEG7QA0H/p1610539399179200] that we weren't creating a ticket to do this work (as I had suggested, so we don't forgot to do it!), we were just going to crack on and get it done. If we maintain the current trajectory it will be forgotten and potential issues will remain within the system. I don't mind either way, but it needs to be done now or recorded so it gets done later.

timabell:
> I was under the impression from the slack conversation [https://dfc-beta.slack.com/archives/G01CEG7QA0H/p1610539399179200] that we weren't creating a ticket to do this work (as I had suggested, so we don't forgot to do it!), we were just going to crack on and get it done. If we maintain the current trajectory it will be forgotten and potential issues will remain within the system. I don't mind either way, but it needs to be done now or recorded so it gets done later.

Is there an issue with this patch that means we shouldn't merge it or can this PR be approved?

timabell:
I don't have the mental bandwidth for the extra analysis on top of the current task. Killing for another day.

jameslawesynergy:
Resurrected and extended following analysis of impact to changes.

The only potential issue we might see is on FAC where an active course has an Archived Venue attached to it, in which case we'll need to fix the data and establish how that came to be.

PR #1874, opened 2021-01-15
PTCD-793 Added Live T Levels Report

jameslawesynergy:


src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetLiveTLevelsReportHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        public async IAsyncEnumerable<LiveTLevelsReportItem> Execute(SqlTransaction transaction, GetLiveTLevelsReport query)
+        {
+            var sql = @$"
+SELECT      p.Ukprn,
+            p.ProviderName,
+            d.Name as TLevelName,
+            v.VenueName,
+            t.StartDate
+FROM        Pttcd.TLevels t with (nolock)
+INNER JOIN  Pttcd.TLevelLocations l with (nolock) ON l.TLevelId = t.TLevelId
```
gunndabad, line 20: Is the `nolock` stuff necessary at this point?

src/Dfc.CourseDirectory.WebV2/Features/HelpdeskDashboard/Dashboard.cshtml
```diff
@@ -15,6 +15,14 @@
                     </ul>
                 </div>
             </div>
+            <div class="govuk-grid-column-one-half">
```
gunndabad, line 18: Need to check for feature flag here?

jameslawesynergy: Good spot. Also included on controller.

src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetLiveTLevelsReportHandler.cs
jameslawesynergy: This is a potentially long running query that could lock the entire t-levels table if we omit the nolock hint for the duration.

gunndabad: Is that the case for snapshot isolation?

jameslawesynergy: I thought we were now REPEATABLE READ?

gunndabad: Reverted to snapshot now by default. From a quick test, snapshot won't block anything if used in a read capacity like above.

src/Dfc.CourseDirectory.WebV2/Features/TLevels/Reporting/TLevelsReportsController.cs
```diff
@@ -0,0 +1,60 @@
...
+        public TLevelsReportsController(IMediator mediator, IClock clock)
+        {
+            _mediator = mediator;
+            _clock = clock;
+        }
+
+        [HttpGet("live-t-levels")]
+        public async Task LiveTLevelsReport()
+        {
+            var results = await _mediator.Send(new LiveTLevelsReport.Query());
```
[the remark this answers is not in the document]
jameslawesynergy: Nice idea. On it's own, perhaps a premature abstraction, but I've reworked the QA reports to leverage it so useful.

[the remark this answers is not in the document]
gunndabad: I don't think it's premature; modifying `HttpResponse` directly in an action is not what the built-in methods do (it potentially stops filters that modify the result after action executing working, for one).

src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetLiveTLevelsReportHandler.cs
jameslawesynergy: What are the chances of that changing again in the future? I'll always lean towards caution on these things.

gunndabad: We've used Snapshot forever apart from for a few weeks (where changing it caused more problems that it solved). Leaving `nolock` in seems fine.

gunndabad (approved):

PR #1878, opened 2021-01-20
Fixed function app startup for FindAnApprenticeshipApi

soharadesign:


gunndabad (approved):
Would be good to nuke src/Dfc.CourseDirectory.FindAnApprenticeshipApi/Properties/launchSettings.json too

PR #1889, opened 2021-01-25
Updated GitVersion config for FAC and FAT

soharadesign:


Dfc.CourseDirectory.FindACourseApi/Resources/AzureDevops/GitVersion.yml
```diff
@@ -1,4 +1,67 @@
 mode: ContinuousDelivery
 branches:
   master:
-    regex: ^main
\ No newline at end of file
+    regex: ^master$|^main$
```
jameslawesynergy, line 4: Isn't master dead?

jameslawesynergy (approved):
Are we currently using GitVersion for CourseDirectory? If so, are we only leveraging the functionality for release?

Suggest we could update pull-request to pick up PTCD-[0-9] prefix in PR title to increment minor version?

Dfc.CourseDirectory.FindACourseApi/Resources/AzureDevops/GitVersion.yml
soharadesign: yeh! it's a generic script and I'm far too lazy to adjust it 👍

PR #1895, opened 2021-01-27
PTCD-585 Fix Empty ApprenticeshipLocation DeliveryModes Function

jameslawesynergy:
Added FixEmptyApprenticeshipLocationDeliveryModes to Functions

### DIFF
--- a/src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/ProcessAllApprenticeships.cs
+++ b/src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/ProcessAllApprenticeships.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq.Expressions;
 using System.Threading.Tasks;
 using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Models;
 using OneOf.Types;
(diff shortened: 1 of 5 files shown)


src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/ProcessAllApprenticeships.cs
```diff
@@ -8,6 +9,7 @@ namespace Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries
 {
     public class ProcessAllApprenticeships : ICosmosDbQuery<None>
     {
+        public Expression<Func<Apprenticeship, bool>> Predicate { get; set; }
```
gunndabad, line 12:
```suggestion
        public Expression<Func<Apprenticeship, bool>> Filter { get; set; }
```

soharadesign (approved):

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/ProcessAllApprenticeships.cs
jameslawesynergy: `Predicate` is the correct term.

soharadesign (approved):

PR #1897, opened 2021-01-27
PTCD-827 Fix checking postcode lookup in FAC search operation

gunndabad:


src/Dfc.CourseDirectory.FindACourseApi/Features/Search/Search.cs
```diff
@@ -382,15 +382,14 @@ private async Task<(float lat, float lng)?> TryGetCoordinatesForPostcode(string
         {
             var result = await _onspdSearchClient.Search(new OnspdSearchQuery() { Postcode = postcode });
 
-            if (result.TotalCount != 0)
+            var doc = result.Items.SingleOrDefault();
```
jameslawesynergy, line 385: *cough* Check affirmatives! *cough* 😄

src/Dfc.CourseDirectory.FindACourseApi/Features/Search/Search.cs
```diff
@@ -382,15 +382,14 @@ private async Task<(float lat, float lng)?> TryGetCoordinatesForPostcode(string
...
+
+            if (doc != null)
```
jameslawesynergy, line 387: Invert `if`

jameslawesynergy (approved):

soharadesign (approved):

src/Dfc.CourseDirectory.Core/Validation/PostcodeHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    public static class PostcodeHelper
+    {
+        public static readonly Regex UkPostcodePattern = new Regex(
+            @"^((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})$",
+            RegexOptions.Compiled);
+
+        public static string NormalizePostcode(string postcode)
+        {
+            if (postcode == null)
```
jameslawesynergy, line 14: `string.IsNullOrWhiteSpace()`?

src/Dfc.CourseDirectory.Core/Validation/PostcodeHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            if (!UkPostcodePattern.IsMatch(postcode))
+            {
+                throw new ArgumentException("Postcode is not valid.", nameof(postcode));
+            }
+
+            // See https://www.bph-postcodes.co.uk/guidetopc.cgi
+
+            var trimmed = postcode.Replace(" ", "").Trim();
+            return trimmed[0..(trimmed.Length - 3)] + " " + trimmed[^3..];
```
jameslawesynergy, line 27: Use string interpolation

src/Dfc.CourseDirectory.Core/Validation/PostcodeHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Dfc.CourseDirectory.Core.Validation
+{
+    public static class PostcodeHelper
+    {
+        public static readonly Regex UkPostcodePattern = new Regex(
+            @"^((([A-Za-z][0-9]{1,2})|(([A-Za-z][A-Ha-hJ-Yj-y][0-9]{1,2})|(([A-Za-z][0-9][A-Za-z])|([A-Za-z][A-Ha-hJ-Yj-y][0-9][A-Za-z]?))))\s?[0-9][A-Za-z]{2})$",
+            RegexOptions.Compiled);
+
+        public static string NormalizePostcode(string postcode)
```
jameslawesynergy, line 12: This feels like a smell. I'm no fan of "Helper" classes as you know and this Normalize function doesn't just normalize, but validates as well.

I suggest doing something similar to the following (not complete implementation):

```
public class PostCode
{
    private string _postCode;

    public PostCode(string postCode)
    {
        _postCode = postCode;
    }
    
    public static bool TryParse(string input, out PostCode postCode)
    {
        // validation, normalization and implementation
    }
    
    public static bool Validate(string postCode)
    {
    }
}
```

This then at least is a useful, descriptive and reusable entity that's available throughout the system.

jameslawesynergy (changes requested):

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+        }
+
+        private Postcode(string value, bool validated)
+        {
+            if (!validated)
+            {
+                if (!ValidateAndNormalize(value, out value))
+                {
+                    throw new ArgumentException("Input is not a valid UK postcode.", nameof(value));
```
jameslawesynergy, line 25: Use `nameof(value)` instead of `"Input"` in exception message.

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
```diff
@@ -0,0 +1,79 @@
...
+                return false;
+            }
+
+            postcode = new Postcode(normalized, validated: true);
+            return true;
+        }
+
+        private static bool ValidateAndNormalize(string input, out string normalized)
+        {
+            if (input == null)
```
jameslawesynergy, line 46: Check `string.IsNullOrWhiteSpace(input)`?

src/Dfc.CourseDirectory.FindACourseApi/Features/Search/Search.cs
```diff
@@ -74,11 +74,11 @@ public class Handler : IRequestHandler<Query, OneOf<ProblemDetails, SearchViewMo
                 };
             }
 
-            var gotPostcode = !string.IsNullOrWhiteSpace(request.Postcode);
+            Postcode postcode = null;
 
-            if (gotPostcode)
+            if (!string.IsNullOrWhiteSpace(request.Postcode))
```
jameslawesynergy, line 79: Unnecessary check for null or whitespace, consider combining for readability.

```suggestion
            if (!Postcode.TryParse(request.Postcode, out var postcode))
```

jameslawesynergy (approved):

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
gunndabad: Parameter name is already part of the exception. The message mirrors `int.Parse()` and `Guid.Parse()`.

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
gunndabad: The Regex check below handles all that. This is only here to avoid the `ArgumentNullException` on `Regex.IsMatch()`

src/Dfc.CourseDirectory.FindACourseApi/Features/Search/Search.cs
gunndabad: It's not unnecessary.

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
jameslawesynergy: `int.Parse()` and `Guid.Parse()` both throw `FormatException` for invalid input data, neither of which use the ctor with an overload for `paramName`. In the case of int.Parse, the parameter name is called `s` and the message refers to "string input", for `Guid.Parse()` the parameter is called `input` and the exception message refers to "input". From what I can see this doesn't mirror their implementations at all.

The variable you're using and what this exception refers to is called `value`, the phrase "Input" means nothing in the context of this execution.

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
jameslawesynergy: Missing an opportunity for an early out without having to invoke an expensive regex?

src/Dfc.CourseDirectory.FindACourseApi/Features/Search/Search.cs
jameslawesynergy: How is it necessary? `Postcode.TryParse(...)` checks this?

jameslawesynergy: If you insist on keeping it this way it should be combined with the child condition to reduce nesting.

gunndabad: It's necessary because postcode parsing (and the subsequent error handling if that fails) should only happen if the request actually specifies a postcode.

src/Dfc.CourseDirectory.Core/Models/Postcode.cs
gunndabad: Modified to use `FormatException`. Given that the only public API that throws this exception takes a single input parameter I am happy the message is clear.

PR #1900, opened 2021-01-28
Better exception messages for missing required Venue fields

timabell:
Reproduce by selecting Buckingham Palace SW1A 1AA which has no `address1` when creating a new venue in the provider facing website.

Triggering the exception:

Add location – Mozilla Firefox_20210128-01-024e3024e3024e3

Add a location – Mozilla Firefox_20210128-01-6d8ef6d8ef6d8ef

Selection_20210128-01-281852818528185

The user won't see this message, they'll just get the generic 500 page, but it'll be easier to understand the logs with a better message. The journey is to be fixed properly at some point so this is a stop-gap.

### DIFF
--- a/src/Dfc.CourseDirectory.Services/Models/Venues/Venue.cs
+++ b/src/Dfc.CourseDirectory.Services/Models/Venues/Venue.cs
@@ -81,22 +81,22 @@ namespace Dfc.CourseDirectory.Services.Models.Venues
 
             if (string.IsNullOrWhiteSpace(venueName))
             {
-                throw new ArgumentException("message", nameof(venueName));
+                throw new ArgumentException("Error creating Venue: Required field is null or whitespace", nameof(venueName));
             }
 
             if (string.IsNullOrWhiteSpace(address1))
             {
-                throw new ArgumentException("message", nameof(address1));
+                throw new ArgumentException("Error creating Venue: Required field is null or whitespace", nameof(address1));
             }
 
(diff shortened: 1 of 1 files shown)


src/Dfc.CourseDirectory.Services/Models/Venues/Venue.cs
```diff
@@ -81,22 +81,22 @@ public class Venue
 
             if (string.IsNullOrWhiteSpace(venueName))
             {
-                throw new ArgumentException("message", nameof(venueName));
+                throw new ArgumentException("Error creating Venue: Required field is null or whitespace", nameof(venueName));
```
jameslawesynergy, line 84: How about `"Null or whitespace {nameof(venueName)}.`"?

jameslawesynergy (changes requested):

src/Dfc.CourseDirectory.Services/Models/Venues/Venue.cs
jameslawesynergy: Or just something that is descriptive of the error.

timabell: The `nameof` is already rolled up into the string by `ArgumentException`, is there a reason for repeating the value in the `message` param?

Selection_20210129-01-831538315383153

jameslawesynergy (approved):

PR #1903, opened 2021-01-29
Filter Out Archived Venues in DataStore.Cosmos Handlers

jameslawesynergy:
As discussed on slack, there is no use case for returning archived Venues

Resurrection of original PR: https://github.com/SkillsFundingAgency/dfc-coursedirectory/pull/1869 and extended following analysis of impact to changes.

The only potential issue we might see is on FAC where an active course has an Archived Venue attached to it, in which case we'll need to fix the data and establish how that came to be.


timabell (approved):
Do we want to ask QA to look at this before merging to flush out any behaviour changes?

jameslawesynergy:
> Do we want to ask QA to look at this before merging to flush out any behaviour changes?

Negative. These changes won't affect "correct" functionality, only highlight bad data which we may need to fix in the future.

PR #1904, opened 2021-02-01
PTCD-830 T Level Detail operation

gunndabad:


src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelDetail.cs
```diff
@@ -0,0 +1,130 @@
...
+        public async Task<OneOf<NotFound, TLevelDetailViewModel>> Handle(Query request, CancellationToken cancellationToken)
+        {
+            var tLevel = await _sqlQueryDispatcher.ExecuteQuery(new GetTLevel() { TLevelId = request.TLevelId });
+
+            if (tLevel == null)
+            {
+                return new NotFound();
+            }
+
+            var getProvider = _cosmosDbQueryDispatcher.ExecuteQuery(
```
jameslawesynergy, line 41: Why do we need to use the Cosmos Provider collection? Is everything not in SQL?

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelDetail.cs
```diff
@@ -0,0 +1,130 @@
...
+                StudyMode = CourseStudyMode.FullTime,
+                TLevelId = tLevel.TLevelId,
+                Website = tLevel.Website,
+                WhatYouCanDoNext = tLevel.WhatYouCanDoNext,
+                WhatYoullLearn = tLevel.WhatYoullLearn,
+                WhoFor = tLevel.WhoFor
+            };
+        }
+
+        private static string EnsureHttpPrefixed(string url) => !string.IsNullOrEmpty(url)
```
jameslawesynergy, line 124: This is duplicated from CourseDetail... what to do with it I wonder? Extension method on string, scoped to FAC? Static within FAC? (not named Helper, 😄) I'm not sure.

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelLocationViewModel.cs
```diff
@@ -1,7 +1,10 @@
-﻿namespace Dfc.CourseDirectory.FindACourseApi.Features.TLevelDetail
+﻿using System;
+
+namespace Dfc.CourseDirectory.FindACourseApi.Features.TLevelDetail
 {
     public class TLevelLocationViewModel
     {
+        public Guid TLevelLocationId { get; set; }
```
jameslawesynergy, line 7: I can't see any updates to the `SELECT` statements in any SqlQueryHandlers to populate this value?

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelDetail.cs
gunndabad: I think this used to be in an `HttpUtil` class. I'm open to naming suggestions but I'm not crazy about making it an extension method on `string`. Could make it an extension method on `Flurl.Url` and use that here (we use Flurl elsewhere already)?

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelDetail.cs
gunndabad: Mainly for symmetry with the CourseRunDetail one. The current SQL queries don't retrieve `Contacts`, which is required here. In this interest of getting this out of the door quickly I did the simplest thing.

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelLocationViewModel.cs
gunndabad: As far as I can see this property is already populated by the query handlers.

jameslawesynergy: Ignore me, I misunderstood that class.

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevelDetail/TLevelDetail.cs
jameslawesynergy: Your call, happy to come back to it later if the solution isn't apparent right now.

jameslawesynergy (approved):

soharadesign (approved):

PR #1912, opened 2021-02-04
PTCD-801 Venues Delete Confirmation Data Integrity

jameslawesynergy:
Created WebV2 Delete Location and added dependent courses, apprenticeships and t levels validation
Refactored View Locations to WebV2
Moved EditVenue to Features/Venues


src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenuesController.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+            _mediator = mediator;
+            _providerContextProvider = providerContextProvider;
+        }
+
+        [HttpGet]
+        public Task<IActionResult> ViewVenues() =>
+            _mediator.SendAndMapResponse<ViewModel, IActionResult>(
+                new Query { ProviderId = _providerContextProvider.GetProviderContext().ProviderInfo.ProviderId },
+                vm => View(vm));
```
timabell, line 24: I'm still in awe of these one-statement controller actions.
A++ would `MediatR` again.

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cs
```diff
@@ -0,0 +1,65 @@
...
+                {
+                    VenueId = r.Id,
+                    VenueName = r.VenueName,
+                    AddressParts = new[]
+                    {
+                        r.AddressLine1,
+                        r.AddressLine2,
+                        r.Town,
+                        r.County
+                    }.Where(part => !string.IsNullOrWhiteSpace(part)).ToArray(),
```
timabell, line 59: this logic feels like it should be pulled out to something else and unit tested, possibly the cosmos venue model?

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cshtml
```diff
@@ -0,0 +1,56 @@
...
+                            <td class="govuk-table__cell" data-testid="venue-postcode">@venue.PostCode</td>
+                            <td class="govuk-table__cell"><a asp-action="Details" asp-controller="EditVenue" asp-route-venueId="@venue.VenueId" class="govuk-link" data-testid="venue-view-link">View<span class="govuk-visually-hidden"> @venue.VenueName</span></a></td>
+                            <td class="govuk-table__cell"><a asp-action="DeleteVenue" asp-controller="DeleteVenue" asp-route-venueId="@venue.VenueId" class="govuk-link" data-testid="venue-delete-link">Delete<span class="govuk-visually-hidden"> @venue.VenueName</span></a></td>
+                        </tr>
+                    }
+                }
+                else
+                {
+                    <tr class="govuk-table__row">
+                        <td class="govuk-table__cell govuk-!-padding-bottom-5 govuk-!-padding-top-5" colspan="5">You have no locations</td>
```
timabell, line 46: this would be the one for them matching `data-testid` for the test commented on elsewhere

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
```diff
@@ -0,0 +1,232 @@
...
+        public async Task<OneOf<NotFound, ViewModel>> Handle(Query request, CancellationToken cancellationToken)
+        {
+            var venue = await _cosmosDbQueryDispatcher.ExecuteQuery(new GetVenueById { VenueId = request.VenueId });
+
+            if (venue == null)
+            {
+                return new NotFound();
+            }
+
+            var getCourses = _cosmosDbQueryDispatcher.ExecuteQuery(new GetAllCoursesForProvider { ProviderUkprn = venue.Ukprn, CourseStatuses = CourseStatus.Live | CourseStatus.Pending | CourseStatus.BulkUploadPending | CourseStatus.BulkUploadReadyToGoLive | CourseStatus.APIPending | CourseStatus.APIReadyToGoLive | CourseStatus.MigrationPending | CourseStatus.MigrationReadyToGoLive });
```
timabell, line 110: needs wrapping because github

timabell: How did you come to that particular list of statuses to check for? Might be worth an explanatory comment.

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
```diff
@@ -0,0 +1,232 @@
...
+            var getApprenticeships = _cosmosDbQueryDispatcher.ExecuteQuery(new GetApprenticeships { Predicate = a => a.ProviderUKPRN == venue.Ukprn && a.ApprenticeshipLocations.Any(al => al.VenueId == request.VenueId) });
```
timabell, line 111: are we going to have efficiency or RU issues with this?

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
timabell: are we going to have efficiency or RU issues with this?

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
```diff
@@ -0,0 +1,232 @@
...
+            var getTLevels = _sqlQueryDispatcher.ExecuteQuery(new SqlQueries.GetTLevelsForProvider { ProviderId = request.ProviderId });
+
+            await Task.WhenAll(getCourses, getApprenticeships, getTLevels);
+
+            return CreateViewModel(venue, request.ProviderId, await getCourses, (await getApprenticeships).Values, await getTLevels);
```
timabell, line 116: I don't understand why these are awaited twice.
I don't see an obvious cache of the result here. Won't it just run it all twice and double the number of requests?

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
```diff
@@ -0,0 +1,232 @@
...
+            var journeyInstance = _journeyInstanceProvider.GetInstance<JourneyModel>();
+            journeyInstance.ThrowIfNotCompleted();
+
+            return Task.FromResult(new DeletedViewModel
+            {
+                VenueName = journeyInstance.State.VenueName
+            });
+        }
+
+        private static ViewModel CreateViewModel(Venue venue, Guid providerId, IEnumerable<Course> courseRuns, IEnumerable<Apprenticeship> apprenticeships, IEnumerable<SqlModels.TLevel> tLevels) => new ViewModel
```
timabell, line 161: wrap because github :cry:

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
```diff
@@ -0,0 +1,232 @@
...
+            AffectedApprenticeships = GetAffectedApprenticeships(apprenticeships, venue.Id),
+            AffectedTLevels = GetAffectedTLevels(tLevels, venue.Id),
+            VenueName = venue.VenueName,
+            AddressParts = new[]
+            {
+                venue.AddressLine1,
+                venue.AddressLine2,
+                venue.Town,
+                venue.County
+            }.Where(part => !string.IsNullOrWhiteSpace(part)).ToArray(),
```
timabell, line 175: repeats logic from elsewhere, stronger argument for pulling out and unit testing, see my other comment

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cshtml
```diff
@@ -0,0 +1,183 @@
...
+                        }
+
+                        <govuk-details>
+                            <govuk-details-summary>@Model.AffectedCourses.Count Course@(Pluralise(Model.AffectedCourses.Count > 1))</govuk-details-summary>
+                            <govuk-details-text>
+                                <p class="govuk-body">The following courses are linked to this location</p>
+                                <ul class="govuk-list govuk-list--bullet">
+                                    @foreach (var affectedCourseRun in Model.AffectedCourses.OrderBy(c => c.CourseName))
+                                    {
+                                        <li data-testId="affected-course-@affectedCourseRun.CourseId"><a asp-controller="CourseSummary" asp-action="Index" asp-route-courseId="@affectedCourseRun.CourseId" asp-route-courseRunId="@affectedCourseRun.CourseRunId" class="govuk-link">@affectedCourseRun.CourseName</a></li>
```
timabell, line 87: wrap because github reivew interface can't, also a little hard to mentally parse like this. suggest breaking before attributes to avoid introducing whitespace to output

timabell:
```suggestion
                                        <li data-testId="affected-course-@affectedCourseRun.CourseId"><a
                                        asp-controller="CourseSummary" asp-action="Index" asp-route-courseId="@affectedCourseRun.CourseId" asp-route-courseRunId="@affectedCourseRun.CourseRunId"
                                          class="govuk-link">@affectedCourseRun.CourseName</a></li>
```
or similar, same for other really long lines in this file

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/DeleteVenueHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+            }
+            catch (DocumentClientException ex)
+                when (ex.StatusCode == HttpStatusCode.NotFound)
+            {
+                return new NotFound();
+            }
+
+            venue.Status = (int)VenueStatus.Archived;
+
+            await client.ReplaceDocumentAsync(documentUri, venue);
```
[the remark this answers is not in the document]
jameslawesynergy: Nope. There's no point in checking for concurrency on a delete operation, as once it's deleted, there's no coming back, so even if an updated slipped in there it wouldn't make a difference.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Venues/ViewVenues/ViewVenuesTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            var request = new HttpRequestMessage(HttpMethod.Get, $"/venues?providerId={providerId}");
+
+            // Act
+            var response = await HttpClient.SendAsync(request);
+
+            // Assert
+            response.StatusCode.Should().Be(StatusCodes.Status200OK);
+
+            var doc = await response.GetDocument();
+            doc.Body.TextContent.Should().Contain("You have no locations");
```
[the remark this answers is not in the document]
jameslawesynergy: I'm not sure what your point is? We use data-testid everywhere in V2 and JG wrote a tag helper to strip these out in release builds: https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/main/src/Dfc.CourseDirectory.WebV2/TagHelpers/RemoveTestDataAttributesTagHelper.cs

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenuesController.cs
jameslawesynergy: I dislike the pattern, but that's one for another day.

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cs
jameslawesynergy: I diagree - it's ViewModel related logic and we shouldn't be sullying a domain/storage model with it.

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cshtml
jameslawesynergy: I don't know what you mean?

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
jameslawesynergy: > needs wrapping because github

I'll wrap it a little, but I won't wrap an `OR` enum value.

> How did you come to that particular list of statuses to check for? Might be worth an explanatory comment.

It's everything but `Archived` and `Deleted`. It's a horrible API but it doesn't require a comment to explain.

> are we going to have efficiency or RU issues with this?

Depends on how much load there is on the site (Cosmos) at any given time. I've discussed this with Cathie and we're happy to push forward on the understanding that it may fail, but the user receiving a 500 error is less bad then them being allowed to perform an action that affects data integrity. If it becomes a serious issue in the future we can limit the number of records returned and provide the user with appropriate messaging.

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
jameslawesynergy: See above comment

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
jameslawesynergy: If you await an already completed `Task`, you get the result. The task is initially awaited by `await.Task.WhenAll(...)` - when that's completed, the tasks are completed and can be awaited again to yield their result.

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
jameslawesynergy: Yeah that's more readable.

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
jameslawesynergy: Same argument from earlier. I diagree - it's ViewModel related logic and we shouldn't be sullying a domain/storage model with it. I'd also advocate maintaining a separation of concerns in View logic across features in this instance as these may diverge in future.

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cshtml
jameslawesynergy: I've wrapped it as sensibly as HTML can be wrapped.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Venues/ViewVenues/ViewVenuesTests.cs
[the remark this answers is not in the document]
timabell: My point is that this isn't using `GetElementByTestId` to find the element containing the expected text and I think it should be.

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenuesController.cs
timabell: I'd love to hear more, blog post? :grin:

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cs
timabell: Because the motivation for this processing isn't clear from the code it's hard for me to judge where the logic to sit. I don't really know *why* this transformation exists. Perhaps you could enlighten me?

Perhaps the viewmodel could expose something specific for the transformation and we could test that?

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cshtml
timabell: counterpart to this comment https://github.com/SkillsFundingAgency/dfc-coursedirectory/pull/1912#discussion_r571988355

timabell: Given the string is being referenced in tests I'd expect to see a `data-testid` in some containing markup. Probably the `td`

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
timabell: > discussed this with Cathie
:+1: 

> "It's everything but Archived and Deleted"

That's knowledge that will be lost to time if it's not in a comment or commit message; and when/if another enum value is added it won't be obvious whether to extend the list with the new value without reverse engineering the original intent. In my view you can help a future soul by adding some more context.

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
timabell: today I learned...
thanks

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cs
jameslawesynergy: > I don't really know _why_ this transformation exists. Perhaps you could enlighten me?

It's a view model - the transformation exists to prepare the data to be displayed in a view.

> Perhaps the viewmodel could expose something specific for the transformation and we could test that?

What's the difference of doing it here or explicitly on the view model? I don't see the value in unit testing this. Notwithstanding, the output of the view, including this data, is covered by existing tests.

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Venues/ViewVenues/ViewVenuesTests.cs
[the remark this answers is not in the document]
jameslawesynergy: I understand you now. Your suggestion actually makes the test more fragile, as the assertion isn't just that the page tells you that information, but that it does so in a specific way. Do we care [for the purpose of this test] how that information is rendered, as long as it is rendered?

src/Dfc.CourseDirectory.WebV2/Features/Venues/DeleteVenue/DeleteVenue.cs
timabell: That's an interesting thought about divergence. I suspect the translation for display would be expected to be consistent across the whole UI, so maybe something more explicitly shared would make sense. Perhaps a question for Westley about the intent.

timabell (approved):
*outstanding* pr, good job :medal_military: :grin:

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cs
timabell: > the transformation exists to prepare the data to be displayed in a view

and a layer deeper, why does it need to be displayed this way as opposed to without any transformation at all? Is it a technical consideration (something to do with model shapes, if so why are they different shapes), or is it a business requirement (if so what is that requirement in plain english and where did it come from).

> What's the difference of doing it here or explicitly on the view model

the point for me is to call it out as a required behaviour across the site, intended for reuse wherever we display this data, and a place to document the motivation for the transformation (which remains unclear to me with the current code).

I hope I'm making some sense

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/Venues/ViewVenues/ViewVenuesTests.cs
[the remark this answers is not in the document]
timabell: I agree it makes the test more fragile, however the point of the ids is so that when you are working in pure markup you can see at a glance that you'll be affecting tests as you modify the markup. Without the `data-testid` you'll be blind to that till you break the test. For me that trade-off is well worth it.

Here's the original article on `data-js` which is the basis for `data-testid` and explains the value in not just hooking non-css things to the existing dom and classes: https://ultimatecourses.com/blog/data-js-selectors-enhancing-html5-development-by-separating-css-from-javascript

src/Dfc.CourseDirectory.WebV2/Features/Venues/ViewVenues/ViewVenues.cs
jameslawesynergy: You're making sense, and I see where you're coming from. However the assumption that all address data throughout the system should be displayed in exactly the same way [hence your desire to standardise it] is perhaps incorrect.

In this particular case, all address fields may not be populated, hence removing the ones that aren't in the view model so as to create less clutter in the view logic. But this logic may not be correct elsewhere on the site, and by creating dependencies across views on formatting logic we introduce a maintenance overhead.

Equally if we ever change the way we want to display addresses, there's a maintenance overhead there too. So basically we're damned if we do, damned if we don't. But in this instance I'm leaning towards maintaining discreet logic for individual views to avoid future headaches and give us maximum flexibility on what is rendered in the view.

timabell: > perhaps incorrect

I think we should do a quick check with the mighty Westley, he'll probably have an opinion on that one way or the other

---

I think I've exhausted my thoughts on the topic, probably to your relief :grin:

gunndabad: We have similar versions of this all over the place:
https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/main/src/Dfc.CourseDirectory.WebV2/SharedViews/Components/ProviderInfoPanelViewComponent.cs#L72-L90
https://github.com/SkillsFundingAgency/dfc-coursedirectory/blob/main/src/Dfc.CourseDirectory.WebV2/Features/ApprenticeshipQA/ProviderApprenticeshipQAInfoPanel.cs#L104-L122

soharadesign (approved):

PR #1915, opened 2021-02-08
PTCD 835 T Levels Canary Feature Flags

jameslawesynergy:
Added CanaryUserFeatureFlagProvider and configured TLevelsBeta flag with admin user to provide TLevels flag

### DIFF
--- a/src/Dfc.CourseDirectory.Core/IFeatureFlagProvider.cs
+++ b/src/Dfc.CourseDirectory.Core/IFeatureFlagProvider.cs
@@ -4,6 +4,6 @@ namespace Dfc.CourseDirectory.Core
 {
     public interface IFeatureFlagProvider
     {
-        ISet<string> GetFeatureFlags();
+        IReadOnlyCollection<string> GetFeatureFlags();
     }
 }
--- a/src/Dfc.CourseDirectory.Web/appsettings.Environment.PreProduction.json
+++ b/src/Dfc.CourseDirectory.Web/appsettings.Environment.PreProduction.json
@@ -3,7 +3,7 @@
     "ApiBaseUri": "https://pp-api.signin.education.gov.uk/",
     "ServiceId": "724D02EB-221E-4638-9E01-9B49B1E67750"
   },
-  "FeatureFlags": "",
+  "FeatureFlags": "TLevelsBeta",
   "GoogleAnalytics": {
     "ClientId": "UA-139898085-3"
   },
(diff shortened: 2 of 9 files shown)


src/Dfc.CourseDirectory.Web/appsettings.Environment.PreProduction.json
```diff
@@ -3,7 +3,7 @@
     "ApiBaseUri": "https://pp-api.signin.education.gov.uk/",
     "ServiceId": "724D02EB-221E-4638-9E01-9B49B1E67750"
   },
-  "FeatureFlags": "",
+  "FeatureFlags": "TLevelsBeta",
```
jameslawesynergy, line 6: As discussed with Steve, enabled in PreProd only for testing in isolation and in an environment where we can confirm that T Levels aren't published by FAC

src/Dfc.CourseDirectory.WebV2/TLevelsBetaCanaryUserFeatureFlagProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+        {
+        }
+
+        protected override ISet<string> ResolveFeatureFlags(AuthenticatedUserInfo user, ISet<string> featureFlags)
+        {
+            if (!featureFlags.Contains(FeatureFlags.TLevels)
+                && featureFlags.Contains(FeatureFlags.TLevelsBeta)
+                && (user?.IsAdmin ?? false))
+            {
+                featureFlags.Add(FeatureFlags.TLevels);
```
gunndabad, line 20: Given the signature of this method I'd expect it to return a new `ISet<T>` not mutate the argument. If mutating the existing instance is expected then I think this method should be `void`.

gunndabad: Perhaps the `featureFlags` argument should be `IReadOnlyCollection<T>` (given we don't have `IReadOnlySet<T>` available to us yet).

jameslawesynergy: I see your point based on the concrete implementation, however I designed the API for this on the basis of producing an `ISet<string>` - the emitted result doesn't necessarily have to derive from the `ISet<string>`, it just so happens to do so in this circumstance.

> Perhaps the featureFlags argument should be IReadOnlyCollection<T> (given we don't have IReadOnlySet<T> available to us yet).

That might be a reasonable approach, I'll look into it.

gunndabad (approved):

src/Dfc.CourseDirectory.Core/IFeatureFlagProvider.cs
```diff
@@ -4,6 +4,6 @@ namespace Dfc.CourseDirectory.Core
 {
     public interface IFeatureFlagProvider
     {
-        ISet<string> GetFeatureFlags();
+        IReadOnlyCollection<string> GetFeatureFlags();
```
jameslawesynergy, line 7: Moved to `IReadOnlyCollection` as discussed with JG - once we're at .NET 5, we can use `IReadOnlySet` instead.

PR #1916, opened 2021-02-08
PTCD-821 Add CostDescription to T Level responses in FAC API

gunndabad:
Populates the `CostDescription` field (and returns `null` for the `Cost` field) for T Level responses on FAC API.

Having this hard-coded in both SQL and C# isn't ideal but we don't currently have a way of sharing these constants across both places.


src/Dfc.CourseDirectory.Database/Pttcd/StoredProcedures/RefreshFindACourseIndexForTLevels.sql
```diff
@@ -23,8 +23,8 @@ BEGIN
 			1 DeliveryMode,  -- Classroom based,
 			CAST(0 AS bit) FlexibleStartDate,
 			t.StartDate,
-			0 Cost,
-			NULL CostDescription,
+			NULL Cost,
+			'T Levels are currently only available to 16-19 year olds.  Contact us for details of other suitable courses.' CostDescription,
```
jameslawesynergy, line 27: Perhaps better to just hard-code this output into the FAC details API rather than stuffing Sql with it? Unless it's required in the search index for anything?

jameslawesynergy (approved):

PR #1925, opened 2021-02-11
Added QualificationLevel to Pttcd.TLevelDefinitions

jameslawesynergy:


src/Dfc.CourseDirectory.Database/Pttcd/Tables/TLevelDefinitions.sql
```diff
@@ -2,6 +2,7 @@
     [TLevelDefinitionId]         UNIQUEIDENTIFIER NOT NULL,
     [FrameworkCode]              INT              NOT NULL,
     [ProgType]                   INT              NOT NULL,
+    [QualificationLevel]         INT     NOT NULL DEFAULT 0,
```
gunndabad, line 5:
```suggestion
    [QualificationLevel]         INT     NOT NULL CONSTRAINT [DF_QualificationLevel] DEFAULT 0,
```

jameslawesynergy: Any particular reason for this syntax over the existing one?

gunndabad: Easier to reference the constraint later if need be.

jameslawesynergy: I don't see why we'd need to reference it later down the line. This should be removed once we have a successful release to prod.

gunndabad: Sure, but if/when we move away from dacpacs and have migration scripts it's considerably easier to remove a constraint when you know its name.

gunndabad (approved):

src/Dfc.CourseDirectory.Database/Pttcd/Tables/TLevelDefinitions.sql
jameslawesynergy: That's a fair point, but we're a long way away from moving from dacpacs - if/when we do move away, we can address any issues at that point. Hopefully this will be long gone by that point.

PR #1932, opened 2021-02-17
PTCD-837 Added TLevels Endpoint to FAC

jameslawesynergy:
Added TLevels endpoint to FAC

### DIFF
--- a/src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
+++ b/src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
@@ -1,0 +1,128 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb;
+using Dfc.CourseDirectory.Core.DataStore.Sql;
+using Dfc.CourseDirectory.Core.DataStore.Sql.Queries;
+using MediatR;
+using CosmosQueries = Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries;
+
+namespace Dfc.CourseDirectory.FindACourseApi.Features.TLevels
+{
+    public class Query : IRequest<ViewModel>
+    {
+    }
+
+    public class ViewModel
+    {
+        public IReadOnlyCollection<TLevelDetailViewModel> TLevels { get; set; }
+    }
+
+    public class Handler : IRequestHandler<Query, ViewModel>
+    {
+        private readonly ISqlQueryDispatcher _sqlQueryDispatcher;
+        private readonly ICosmosDbQueryDispatcher _cosmosDbQueryDispatcher;
+
+        public Handler(ISqlQueryDispatcher sqlQueryDispatcher, ICosmosDbQueryDispatcher cosmosDbQueryDispatcher)
+        {
+            _sqlQueryDispatcher = sqlQueryDispatcher;
+            _cosmosDbQueryDispatcher = cosmosDbQueryDispatcher;
+        }
+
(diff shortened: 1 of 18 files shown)


src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
```diff
@@ -0,0 +1,127 @@
...
+
+        public Handler(ISqlQueryDispatcher sqlQueryDispatcher, ICosmosDbQueryDispatcher cosmosDbQueryDispatcher)
+        {
+            _sqlQueryDispatcher = sqlQueryDispatcher;
+            _cosmosDbQueryDispatcher = cosmosDbQueryDispatcher;
+        }
+
+        public async Task<ViewModel> Handle(Query request, CancellationToken cancellationToken)
+        {
+            var tLevels = await _sqlQueryDispatcher.ExecuteQuery(new GetTLevels());
```
jameslawesynergy, line 36: I'm nervous about the performance of this approach once we begin to see large number of T Levels available within the system.

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
```diff
@@ -0,0 +1,127 @@
...
+            var sqlProviders = await getSqlProviders;
+            var venues = await getVenues;
+
+            var feChoices = await _cosmosDbQueryDispatcher.ExecuteQuery(new CosmosQueries.GetFeChoicesByProviderUkprns { ProviderUkprns = providers.Values.Select(p => p.Ukprn) });
+
+            return new ViewModel
+            {
+                TLevels = tLevels.Select(t =>
+                {
+                    var provider = providers[t.ProviderId];
```
jameslawesynergy, line 61: Pondering: Should this blow up if we have invalid data? Or should we continue to serve up what valid data we have?

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
gunndabad: Agreed - we could look into the `IAsyncEnumerable` from DB -> JSON in future but for a first pass this seems fine.

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
gunndabad: Seems fine as-is.

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            var providerIds = tLevels.Select(t => t.ProviderId).Distinct();
+
+            var getProviders = _cosmosDbQueryDispatcher.ExecuteQuery(
+                new CosmosQueries.GetProvidersByIds { ProviderIds = providerIds });
+
+            var getSqlProviders = _sqlQueryDispatcher.ExecuteQuery(
+                new GetProvidersByIds { ProviderIds = providerIds });
+
+            var getVenues = _sqlQueryDispatcher.ExecuteQuery(
```
gunndabad, line 46: I wonder whether we should add `Venue` onto `TLevelLocation` at some point since all these queries are already going to that table and we often need to lookup the `Venue` for each location. Probably one for the future given the number of places that would need changing.

gunndabad (approved):

src/Dfc.CourseDirectory.FindACourseApi/Features/TLevels/TLevels.cs
jameslawesynergy: That could be useful moving forward, but perhaps out of scope for this.

PR #1933, opened 2021-02-17
Don't show the admin sub-nav on Provider Search page

gunndabad:


jameslawesynergy (changes requested):
I left it working like this deliberately. That way the user can go back to the provider they were currently viewing without having to search again.

gunndabad:
> I left it working like this deliberately. That way the user can go back to the provider they were currently viewing without having to search again. It seemed like an enhancement.

I've checked with the designers - they don't want it there.

jameslawesynergy (approved):

PR #1934, opened 2021-02-17
Remove unused provider search fields

timabell:
No references to these in the code.


jameslawesynergy (changes requested):
They're there because they're searchable fields as per the search index specification.

timabell:
Could you expand on that @jameslawesynergy , I know very little about azure search.

jameslawesynergy (approved):

timabell:
Short answer - these were here to keep parity with contract defined here: https://github.com/SkillsFundingAgency/dfc-providerportal-ukrlp/blob/61d701337d00bb43333c1203d96295a317de85c9/Resources/AzureSearch/SearchIndex.json#L34-L37

Trimming down in this case to reduce the amount of things we have to think about. This is an internal contract (we control both sides).

PR #1935, opened 2021-02-17
PTCD-676 Replace LoqateAddressSearchService with GetAddressAddressSearchService

jameslawesynergy:
Added GetAddress api config values into Arm Templates
Updated GetAddressSearchService to use configured ApiUrl
Updated IAddressSearchService registration to use GetAddressSearchService

### DIFF
--- a/Dfc.CourseDirectory/Resources/ArmTemplates/parameters.json
+++ b/Dfc.CourseDirectory/Resources/ArmTemplates/parameters.json
@@ -14,10 +14,10 @@
     "postcodeApiKey": {
       "value": "__postcodeApiKey__"
     },
-    "getaddressApiUrl":{
+    "getAddressApiUrl":{
       "value": "__getaddressApiUrl__"
     },
-    "getaddressApiKey":{
+    "getAddressApiKey":{
       "value": "__getaddressApiKey__"
     },
     "venueApiUrl": {
--- a/Dfc.CourseDirectory/Resources/ArmTemplates/template.json
+++ b/Dfc.CourseDirectory/Resources/ArmTemplates/template.json
@@ -26,13 +26,13 @@
         "description": "API host key (code) needed to access the postcode search"
       }
     },
-    "getaddressApiUrl": {
+    "getAddressApiUrl": {
       "type": "string",
       "metadata": {
         "description": "API URL for the GetAddress postcode search"
       }
     },
-    "getaddressApiKey": {
+    "getAddressApiKey": {
       "type": "securestring",
       "metadata": {
         "description": "API host key (code) needed to access the GetAddress postcode search"
@@ -384,14 +384,6 @@
                 "value": "[parameters('larsApiKey')]"
               },
               {
-                "name": "GetAddressSettings__ApiUrl",
-                "value": "[parameters('getaddressApiUrl')]"
-              },
-              {
-                "name": "GetAddressSettings__ApiKey",
-                "value": "[parameters('getaddressApiKey')]"
-              },
-              {
                 "name": "LarsSearchSettings__ApiVersion",
                 "value": "2017-11-11"
               },
@@ -418,6 +410,14 @@
               {
                 "name": "PostCodeSearchSettings__Key",
                 "value": "[parameters('postcodeApiKey')]"
+              },
+              {
+                "name": "GetAddressSettings__ApiUrl",
+                "value": "[parameters('getAddressApiUrl')]"
+              },
+              {
+                "name": "GetAddressSettings__ApiKey",
+                "value": "[parameters('getAddressApiKey')]"
               },
               {
                 "name": "VenueServiceSettings__ApiUrl",
--- a/Dfc.CourseDirectory/Resources/ArmTemplates/test-parameters.json
+++ b/Dfc.CourseDirectory/Resources/ArmTemplates/test-parameters.json
@@ -14,10 +14,10 @@
     "postcodeApiKey": {
       "value": "not-a-real-key"
     },
-    "getaddressApiUrl":{
+    "getAddressApiUrl":{
       "value": "https://my.api.net/api/postcode?"
     },
-    "getaddressApiKey":{
+    "getAddressApiKey":{
       "value": "not-a-real-key"
     },
     "venueApiUrl": {
@@ -137,7 +137,7 @@
     "providerPortalCosmosDbConnectionString": {
       "value": "https://not-a-real-account.documents.azure.com/"
     },
-    "providerPortalApplicationInsights":{
+    "providerPortalApplicationInsights": {
       "value": "dfc-dev-prov-as-ai"
     }
   }
--- a/src/Dfc.CourseDirectory.Web/appsettings.json
+++ b/src/Dfc.CourseDirectory.Web/appsettings.json
@@ -79,5 +79,9 @@
     "RetrieveAddressBaseUrl": "https://services.postcodeanywhere.co.uk/PostcodeAnywhere/Interactive/RetrieveById/v1.30/json3.ws?"
   },
   "ProviderAzureSearchIndexName": "provider",
-  "LarsAzureSearchIndexName": "lars"
+  "LarsAzureSearchIndexName": "lars",
+  "GetAddressSettings": {
+    "UseGetAddress": true,
+    "ApiUrl": "https://api.getaddress.io/find/{0}"
+  }
 }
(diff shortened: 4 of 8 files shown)


Dfc.CourseDirectory/Resources/ArmTemplates/test-parameters.json
```diff
@@ -14,6 +14,12 @@
     "postcodeApiKey": {
       "value": "not-a-real-key"
     },
+    "getAddressApiUrl": {
+      "value": "https://my.api.net/getaddress"
```
timabell, line 18: `api.net` is a registered domain https://www.whois.com/whois/api.net I suggest sticking to reserved domains (and their subdomains of course) for tests such as `example.org`
https://www.iana.org/domains/reserved

Dfc.CourseDirectory/Resources/ArmTemplates/template.json
```diff
@@ -26,6 +26,18 @@
         "description": "API host key (code) needed to access the postcode search"
       }
     },
+    "getAddressApiUrl": {
+      "type": "string",
+      "metadata": {
+        "description": "API URL for GetAddress postcode search"
```
timabell, line 32: Does this allow a c# format placeholder for postcode? If so this might be a good place to document the expected format.

tests/Dfc.CourseDirectory.WebV2.Tests/AddressSearch/GetAddressAddressSearchServiceTests.cs
```diff
@@ -21,8 +21,8 @@ public async Task SearchByPostcode_WithValidRequest_ReturnsParsedResults()
             new HttpRequestInterceptionBuilder()
                 .Requests()
                 .ForHttps()
-                .ForHost("api.getaddress.io")
-                .ForPath("find/XX2 00X")
+                .ForHost("testing.com")
```
timabell, line 24:
```suggestion
                .ForHost("example.com")
```
https://www.whois.com/whois/testing.com is a real registered domain
vs https://www.whois.com/whois/example.com which is reserved for test usage etc.
Same for all the others below, I won't repeat the comment.

timabell (commented):
The changes look good to me, however it's not clear to me how the changes achieve what the title of the PR says. There's no mention of `LoquateAddressSearchService` in the patch.

Dfc.CourseDirectory/Resources/ArmTemplates/test-parameters.json
jameslawesynergy: I didn't create the file or the convention so this isn't really helpful in the context of this review.

tests/Dfc.CourseDirectory.WebV2.Tests/AddressSearch/GetAddressAddressSearchServiceTests.cs
jameslawesynergy: Why does it matter in a closed unit test?

Dfc.CourseDirectory/Resources/ArmTemplates/template.json
jameslawesynergy: A pointless exercise I expect. Someone is never going to refer to this a source of documentation having not already looked at the self documenting code.

Dfc.CourseDirectory/Resources/ArmTemplates/parameters.json
```diff
@@ -14,6 +14,12 @@
     "postcodeApiKey": {
       "value": "__postcodeApiKey__"
     },
+    "getAddressApiUrl": {
```
gunndabad, line 17: Why is the URL in the ARM template?

jameslawesynergy: The Loqate one was, but it wasn't used by the implementation. Typically it's good practice to have these values configurable so if they need to change they can be updated without the need for a new release, equally it gives us the option to use a different endpoint for different environments should the need arise.

timabell (approved):

Dfc.CourseDirectory/Resources/ArmTemplates/parameters.json
gunndabad: So it can live in `appsettings.json`

tests/Dfc.CourseDirectory.WebV2.Tests/AddressSearch/GetAddressAddressSearchServiceTests.cs
timabell: Two reasons I can think of:

1. Eliminate the risk of tests accidentally reaching out to real servers at some point in future (you never know what other people will do with your carefully crafted code in the future)
2. Eliminate the risk that the arbitrary mention of someone else's domain becoming problematic for non-technical reasons in the future (for example if the domain you have used becomes part of a malware system, or has some massive public relations disaster).

Given that there's no downside to sticking to the reserved domains I see no reason not to do so as a matter of habit.

Dfc.CourseDirectory/Resources/ArmTemplates/template.json
timabell: > "A pointless exercise I expect. "

I'm not sure I appreciate your tone.

> "Someone is never going to refer to this a source of documentation having not already looked at the self documenting code."

This is specifically documentation for people configuring the service who may not be the current team and may very well not want to have to read all the source code to find out that there's a `string.format` in there somewhere; if they even have the knowledge of c# to do so.

Personally out of empathy for future self/others I would document that it accepts `https://..../.../{postcode}/...`, if I understand the code correctly. Whether you bother is your call, it's just a suggestion.

jameslawesynergy: On reflection that's a fair point, will update.

Dfc.CourseDirectory/Resources/ArmTemplates/test-parameters.json
jameslawesynergy: I didn't create the convention that already exists for this file. Please feel free to update it.

Dfc.CourseDirectory/Resources/ArmTemplates/parameters.json
jameslawesynergy: Potentially, yes.

tests/Dfc.CourseDirectory.WebV2.Tests/AddressSearch/GetAddressAddressSearchServiceTests.cs
jameslawesynergy: Agreed.

Dfc.CourseDirectory/Resources/ArmTemplates/template.json
jameslawesynergy: > I'm not sure I appreciate your tone.

If nobody ever looks at it again, it's a pointless exercise.

> if they even have the knowledge of c# to do so

They'll need to know about the C# code to understand the string formatting, so the comment is useless without the context of the code.

timabell: > If nobody ever looks at it again, it's a pointless exercise.

consider it feedback about the way that came across in response to my efforts to improve our contributions to the codebase. "pointless" is pejorative.
 
> They'll need to know about the C# code to understand the string formatting, so the comment is useless without the context of the code.

I think a simple placeholder replacement could be understood without knowing all of c#, but that aside if you don't feel it's worth the effort to alter this that is your call. Feel free to resolve the comment and move on

src/Dfc.CourseDirectory.Web/appsettings.json
```diff
@@ -79,5 +79,9 @@
     "RetrieveAddressBaseUrl": "https://services.postcodeanywhere.co.uk/PostcodeAnywhere/Interactive/RetrieveById/v1.30/json3.ws?"
   },
   "ProviderAzureSearchIndexName": "provider",
-  "LarsAzureSearchIndexName": "lars"
+  "LarsAzureSearchIndexName": "lars",
+  "GetAddressSettings": {
+    "UseGetAddress": true,
```
jameslawesynergy, line 84: Assuming this can be overridden simply by setting environment variable `GetAddressSettings::UseGetAddress` to `false`?

gunndabad (approved):

PR #1940, opened 2021-02-22
PTCD-804 Provider Type Report

jameslawesynergy:


src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderTypeReportHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+            var sql = @$"
+SELECT      p.Ukprn as ProviderUkprn,
+            p.ProviderName,
+            p.ProviderType,
+            p.ProviderStatus,
+            p.UkrlpProviderStatusDescription,
+            (
+                SELECT      COUNT(*)
+                FROM        Pttcd.Courses c with (nolock)
+                INNER JOIN  Pttcd.CourseRuns cr with (nolock) ON cr.CourseId = c.CourseId
```
gunndabad, line 22: Could eliminate the `CourseRuns` and `ApprenticeshipLocations` joins here by doing something like `(c.CourseStatus & ${(int)CourseStatus.Live}) <> 0`

gunndabad (approved):

src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetProviderTypeReportHandler.cs
jameslawesynergy: Can't for `CourseRuns`, as the "count of courses", is actually the "count of course runs".

As for apprenticeships, that's possible, but I was led to believe that this field was typically unreliable given it's a computed bitwise.

gunndabad: >Can't for CourseRuns, as the "count of courses", is actually the "count of course runs".
Yuk. Fair enough.

The apprenticeship should be reliable and we certainly assuming it is elsewhere. If courses is staying as-is then being consistent makes sense too.

PR #1942, opened 2021-02-23
Add UKRLP Town and County to ProviderContact

jameslawesynergy:
Added address Town and County with mapping from UKRLP to Cosmos and SQL

PR #1943, opened 2021-02-23
Provider Search Index Use ProviderContact Town

jameslawesynergy:
Updated vwProviderSearchIndex to use ProviderContact Town if available

### DIFF
--- a/src/Dfc.CourseDirectory.Database/Pttcd/Views/vwProviderSearchIndex.sql
+++ b/src/Dfc.CourseDirectory.Database/Pttcd/Views/vwProviderSearchIndex.sql
@@ -1,19 +1,19 @@
-create
-    --or alter -- not supported by dacpac
-    view Pttcd.vwProviderSearchIndex
-as
-select
-    p.ProviderId,
-    p.Ukprn,
-    p.ProviderName,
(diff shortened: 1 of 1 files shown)


soharadesign (approved):

timabell (approved):

src/Dfc.CourseDirectory.Database/Pttcd/Views/vwProviderSearchIndex.sql
```diff
@@ -1,19 +1,19 @@
...
-) p_contact
+CREATE VIEW [Pttcd].[vwProviderSearchIndex]
+AS
+SELECT      p.ProviderId,
+            p.Ukprn,
+            p.ProviderName,
+            p.ProviderStatus,
+            p.UkrlpProviderStatusDescription,
+            pc.AddressPostcode Postcode,
+            ISNULL(pc.AddressPostTown, pc.AddressItems) Town,
```
timabell, line 9: If we've now fixed the source of the missing data in #1942 do we still need the `isnull` fallback to `AddressItems` or can we always use `AddressPostTown`. I *think* it'll be fine without the fallback.

PR #1948, opened 2021-02-26
PTCD-852 Fix FAC T Levels Endpoints ProviderAddress Data

jameslawesynergy:
Updated Providers SqlDataSync to fall back to ProviderContact.AddressItems for Town and County
Added ConcatAddressLines to FAC ViewModelFormatting and updated TLevels endpoints ProviderAddress mapping


timabell (approved):

soharadesign (approved):

jameslawesynergy:
Thoughts - should we extend this update to the CouseDetails endpoint which uses a similar schema? If we do, should we engage with the live team as the data provided will change substantially.

PR #1952, opened 2021-03-01
PTCD-823 Revised Add Venue journey

gunndabad:


src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/CreateVenue.cs
```diff
@@ -1,16 +1,16 @@
...
 {
     public class CreateVenue : ICosmosDbQuery<Success>
     {
         public Guid VenueId { get; set; }
         public int ProviderUkprn { get; set; }
-        public string VenueName { get; set; }
+        public string Name { get; set; }
         public string Email { get; set; }
-        public string Telephone { get; set; }
+        public string PhoneNumber { get; set; }
```
jameslawesynergy, line 13: This is `Telephone` in the storage model. Surely we should maintain consistency across the domain with the storage model taking precedence?

gunndabad: This was changed to align with the other models (like in the Update flow). The current model is a total mess so I sided with the VM in this case.

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/CreateVenueHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+                AddressLine1 = request.AddressLine1,
+                AddressLine2 = request.AddressLine2,
+                Town = request.Town,
+                County = request.County,
+                Postcode = request.Postcode,
+                Latitude = request.Latitude,
+                Longitude = request.Longitude,
+                CreatedBy = request.CreatedBy.Email,
+                CreatedDate = request.CreatedDate,
+                UpdatedBy = request.CreatedBy.Email,
```
jameslawesynergy, line 44: I thought we were using `UserId` for this moving forward?

gunndabad: That hasn't been changed for any other Venue journeys yet as far as I know and I wanted to keep consistent.

src/Dfc.CourseDirectory.Web/Controllers/ApprenticeshipsController.cs
```diff
@@ -881,10 +881,19 @@ public IActionResult DeleteDeliveryOption(DeleteDeliveryOptionViewModel model)
         [Authorize]
         public IActionResult AddNewVenue(ApprenticeshipLocationType type)
         {
-            _session.SetString("Option", type == ApprenticeshipLocationType.ClassroomBasedAndEmployerBased ?
-                "AddNewVenueForApprenticeshipsCombined" :"AddNewVenueForApprenticeships");
+            var returnUrl = type == ApprenticeshipLocationType.ClassroomBasedAndEmployerBased ?
+                Url.Action("DeliveryOptionsCombined", "Apprenticeships", new
+                {
+                    message = "",
```
jameslawesynergy, line 887: Not sure why we're using `message` to build the url, do we need to provide it, or just because legacy? If it's needed, use `string.Empty`.

gunndabad: It was ported directly from legacy stuff, I believe. I wanted minimal changes in these legacy journeys.

src/Dfc.CourseDirectory.Web/Views/AddCourse/AddCourseDetails.cshtml
```diff
@@ -37,42 +37,16 @@
...
-        if (!string.IsNullOrWhiteSpace(newVenueObject.County))
-        {
-            message += string.Concat(@newVenueObject.County, ", ");
-        }
-        if (!string.IsNullOrWhiteSpace(newVenueObject.PostCode))
-        {
-            message += newVenueObject.PostCode;
-
-        }
+    var newVenueDescription = TempData[Dfc.CourseDirectory.WebV2.TempDataKeys.AddedVenueDescription] as string;
```
jameslawesynergy, line 40: Was there a requirement to keep this notification stuff in there? (I thought this was a pattern we were moving away from?)

gunndabad: Yeh these are staying as-is for now.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace Dfc.CourseDirectory.WebV2.Features.Venues.AddVenue
+{
+    [Route("venues/add")]
+    [JourneyMetadata(
+        journeyName: "AddVenue",
+        stateType: typeof(AddVenueJourneyModel),
+        appendUniqueKey: true,
+        requestDataKeys: "providerId?")]
```
jameslawesynergy, line 15: Why is this param suffixed with a `?`? Is that an intentional behaviour? Is there a better way to express the meaning if so?

gunndabad: It makes it optional (so it will work when `providerId` isn't present, i.e. for non-admins). That's the only way of expressing that currently. Open to suggestions for how to make it more readable.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+    [RequireProviderContext]
+    public class AddVenueController : Controller
+    {
+        private const string ReturnUrlJourneyInstancePropertyKey = "returnUrl";
+
+        private readonly IMediator _mediator;
+        private readonly JourneyInstanceProvider _journeyInstanceProvider;
+        private readonly IProviderContextProvider _providerContextProvider;
+        private JourneyInstance<AddVenueJourneyModel> _journeyInstance;
```
jameslawesynergy, line 24: I'm not a fan of declaring `JourneyInstance` at the class level. While it's used across multiple methods in the class, it's better suited to being scoped as a local variable as that's its true intent.

In other words, it doesn't **need** to be here, so it shouldn't be.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            _journeyInstance = _journeyInstanceProvider.GetOrCreateInstance(
+                () => new AddVenueJourneyModel(),
+                new PropertiesBuilder()
+                    .Add(ReturnUrlJourneyInstancePropertyKey, returnUrl)
+                    .Build());
+
+            if (!_journeyInstanceProvider.IsCurrentInstance(_journeyInstance))
+            {
+                return RedirectToAction()
```
jameslawesynergy, line 48: Surely we don't want to redirect to itself here, we want to show a meaningful error, perhaps `BadRequest`?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+                    .WithProviderContext(_providerContextProvider.GetProviderContext())
+                    .WithJourneyInstanceUniqueKey(_journeyInstance);
+            }
+
+            return View(
+                nameof(PostcodeSearch),
+                new PostcodeSearch.SearchCommand()
```
jameslawesynergy, line 55: Why are you passing a "Command" to a view? This doesn't make sense to me in the context of the CQRS pattern that's implemented into the existing system.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+                });
+        }
+
+        [HttpPost("cancel")]
+        [RequireJourneyInstance]
+        public IActionResult Cancel()
+        {
+            _journeyInstance.Delete();
+
+            if (_journeyInstance.Properties.TryGetValue(ReturnUrlJourneyInstancePropertyKey, out var returnUrlObj) &&
```
jameslawesynergy, line 67: Accepted practice when breaking lines for conditions is to prefix with the operator.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: What else would we name it?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+        public async Task<IActionResult> PostcodeSearch(PostcodeSearch.SearchCommand command) =>
+            await _mediator.SendAndMapResponse(
+                command,
+                result => result.Match(
+                    errors => this.ViewFromErrors(errors),
+                    vm => View("PostcodeSearchResults", vm)));
+
+        [HttpPost("select-postcode")]
+        [RequireJourneyInstance]
+        public async Task<IActionResult> SelectPostcode(PostcodeSearch.SelectCommand command, [FromQuery] string postcode)
```
jameslawesynergy, line 89: Why is the `postcode` param not included in the command?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: I disagree that it's 'accepted practice' but I will change it.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: One is bound from the body and the other from the query param. I can't remember if there was a reason why the latter is from the query param, other than consistency with the earlier pages. I'll look at if it's easy to change.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+                                _journeyInstance.State.Name,
+                                _journeyInstance.State.AddressLine1,
+                                _journeyInstance.State.AddressLine2,
+                                _journeyInstance.State.Town,
+                                _journeyInstance.State.County,
+                                _journeyInstance.State.Postcode
+                            }
+                            .Where(l => !string.IsNullOrWhiteSpace(l))));
+
+                    if (_journeyInstance.Properties.TryGetValue(ReturnUrlJourneyInstancePropertyKey, out var returnUrlObj) &&
```
jameslawesynergy, line 172: As before re operator suffix

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+                        .WithProviderContext(_providerContextProvider.GetProviderContext())
+                        .WithJourneyInstanceUniqueKey(_journeyInstance);
+                });
+
+        [HttpGet("success")]
+        [RequireJourneyInstance]
+        public async Task<IActionResult> Published()
+        {
+            var query = new Published.Query();
+            return await _mediator.SendAndMapResponse(query, vm => View(vm));
```
jameslawesynergy, line 190: No need to await this! (or many of the other Actions in this class, but this is the obvious one that can be made more succinct)

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cs
```diff
@@ -0,0 +1,116 @@
...
+            var validationResult = await validator.ValidateAsync(request);
+
+            if (!validationResult.IsValid)
+            {
+                return new ModelWithErrors<Command>(request, validationResult);
+            }
+
+            var onspdSearchResult = await _onspdSearchClient.Search(new OnspdSearchQuery() { Postcode = request.Postcode });
+
+            if (onspdSearchResult.Items.Count == 0)
```
jameslawesynergy, line 71:
```suggestion
            if (!onspdSearchResult.Items.Any())
```

gunndabad: Why use something less efficient?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                validationResult = new ValidationResult(new[]
+                {
+                    new ValidationFailure(nameof(request.Postcode), "Enter a valid postcode")
+                });
+
+                return new ModelWithErrors<Command>(request, validationResult);
+            }
+
+            var onspdPostcodeRecord = onspdSearchResult.Items.Single();
```
jameslawesynergy, line 81: In fact, you could move this above the previous condition and check for null to determine results returns

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: Can't just `return` here - the `Task<IViewResult>` isn't assignable to `Task<IActionResult>`

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cs
```diff
@@ -0,0 +1,116 @@
...
+                state.Latitude = onspdPostcodeRecord.Record.lat;
+                state.Longitude = onspdPostcodeRecord.Record.@long;
+                state.AddressIsOutsideOfEngland = onspdPostcodeRecord.Record.Country != "England";
+                state.ValidStages |= AddVenueCompletedStages.Address;
+            });
+
+            return new Success();
+        }
+
+        private void ThrowIfFlowStateNotValid()
```
jameslawesynergy, line 99: Unnecessary abstraction

gunndabad: It's to be consistent with other pages.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cs
```diff
@@ -0,0 +1,116 @@
...
+        public string Town { get; set; }
+        public string County { get; set; }
+        public string Postcode { get; set; }
+    }
+
+    public class Handler :
+        IRequestHandler<Query, Command>,
+        IRequestHandler<Command, OneOf<ModelWithErrors<Command>, Success>>
+    {
+        private readonly JourneyInstance<AddVenueJourneyModel> _journeyInstance;
```
jameslawesynergy, line 34: As comments from Controller, not on a fan of declaring the instance at this level as it doesn't need to be here. Better suited to a scoped variable.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
```diff
@@ -0,0 +1,193 @@
...
+                    success => RedirectToAction(fromPublishPage == true ? nameof(CheckAndPublish) : nameof(Details))
+                        .WithProviderContext(_providerContextProvider.GetProviderContext())
+                        .WithJourneyInstanceUniqueKey(_journeyInstance)));
+
+        [HttpGet("details")]
+        [RequireJourneyInstance]
+        public async Task<IActionResult> Details([FromQuery] bool? fromPublishPage) =>
+            await _mediator.SendAndMapResponse(
+                new Details.Query(),
+                vm => View(vm).WithViewData("FromPublishPage", fromPublishPage));
```
jameslawesynergy, line 125: Should we consider adding this value to the view model as opposed to using the ViewBag?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cshtml
```diff
@@ -0,0 +1,46 @@
...
+
+            <govuk-button type="submit" class="govuk-!-margin-right-3">Continue</govuk-button>
+
+            @if (ViewBag.FromPublishPage == true)
+            {
+                <govuk-button asp-action="CheckAndPublish" asp-all-route-data="@RouteValuesHelper.FromQueryString()" class="govuk-button--secondary">Cancel</govuk-button>
+            }
+            else
+            {
+                <govuk-button type="submit" asp-action="Cancel" asp-all-route-data="@RouteValuesHelper.FromQueryString()" formmethod="post" class="govuk-button--secondary">Cancel</govuk-button>
```
jameslawesynergy, line 42: Why is this a submit button?

gunndabad: Because it's a `POST` (as it mutates state)

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cs
```diff
@@ -0,0 +1,130 @@
...
+            JourneyInstanceProvider journeyInstanceProvider,
+            ICurrentUserProvider currentUserProvider,
+            IClock clock,
+            IProviderContextProvider providerContextProvider)
+        {
+            _cosmosDbQueryDispatcher = cosmosDbQueryDispatcher;
+            _currentUserProvider = currentUserProvider;
+            _clock = clock;
+            _providerContextProvider = providerContextProvider;
+            _journeyInstance = journeyInstanceProvider.GetInstance<AddVenueJourneyModel>();
```
jameslawesynergy, line 55: As previous comments, this should be scoped locally.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        public Task<ViewModel> Handle(Query request, CancellationToken cancellationToken)
+        {
+            ThrowIfFlowStateNotValid();
+
+            var state = _journeyInstance.State;
```
jameslawesynergy, line 62: Any particular reason for this variable?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cs
```diff
@@ -0,0 +1,130 @@
...
+
+            var addressParts = new[]
```
jameslawesynergy, line 64: Any particular reason for this variable?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cshtml
```diff
@@ -0,0 +1,58 @@
+﻿@model Dfc.CourseDirectory.WebV2.Features.Venues.AddVenue.CheckAndPublish.ViewModel
+@{
+    Layout = "_V2LayoutProviderContext";
+    ViewContext.SetLayoutData(title: "Add location");
+}
+
+@if (Model.NewAddressIsOutsideOfEngland)
+{
+    <pttcd-notification title="Non England address" data-testid="outsideofengland-notification">
+        You are adding an address that is outside of England. Your course for this address will not appear in any search results.
```
jameslawesynergy, line 10: Is that really the copy that's been supplied for this error message?

"Your course for this address will not appear in any search results." is poorly worded, particularly the use of "course".

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cshtml
```diff
@@ -0,0 +1,58 @@
...
+                </govuk-summary-list-row>
+                <govuk-summary-list-row>
+                    <govuk-summary-list-row-key>Website</govuk-summary-list-row-key>
+                    <govuk-summary-list-row-value>@Model.Website</govuk-summary-list-row-value>
+                    <govuk-summary-list-row-action asp-action="Details" asp-all-route-data="@RouteValuesHelper.FromQueryString()" asp-route-fromPublishPage="true" asp-fragment="@Html.IdFor(m => m.Website)" visually-hidden-text="website">Change</govuk-summary-list-row-action>
+                </govuk-summary-list-row>
+            </govuk-summary-list>
+
+            <govuk-button type="submit" class="govuk-!-margin-right-3">Save location</govuk-button>
+            <govuk-button type="submit" asp-action="Cancel" asp-all-route-data="@RouteValuesHelper.FromQueryString()" formmethod="post" class="govuk-button--secondary">Cancel</govuk-button>
```
jameslawesynergy, line 55: Should this be a submit button? Surely we just want a text link?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Details.cs
```diff
@@ -0,0 +1,136 @@
...
+    public class ViewModel : Command
+    {
+        public IReadOnlyCollection<string> AddressParts { get; set; }
+    }
+
+    public class Handler :
+        IRequestHandler<Query, ViewModel>,
+        IRequestHandler<Command, OneOf<ModelWithErrors<ViewModel>, Success>>
+    {
+        private readonly JourneyInstance<AddVenueJourneyModel> _journeyInstance;
```
jameslawesynergy, line 40: See previous comments

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Details.cshtml
```diff
@@ -0,0 +1,53 @@
...
+
+            <p class="govuk-body-l">This information will be visible when linked with the courses you provide.</p>
+
+            <govuk-input asp-for="Name" input-class="govuk-!-width-two-thirds">
+                <govuk-input-label class="govuk-label--s">Location name</govuk-input-label>
+            </govuk-input>
+
+            <h2 class="govuk-heading-s govuk-!-margin-bottom-1">Address</h2>
+            <p class="govuk-body">
+
```
jameslawesynergy, line 22: -ws

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Details.cshtml
```diff
@@ -0,0 +1,53 @@
...
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+        <form asp-action="Details" asp-all-route-data="RouteValuesHelper.FromQueryString()">
+            <h1 class="govuk-heading-xl">
+                Add location
+            </h1>
+
+            <p class="govuk-body-l">This information will be visible when linked with the courses you provide.</p>
```
jameslawesynergy, line 14: Copy issue... not just courses

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+        public string Postcode { get; set; }
+    }
+
+    public class Handler :
+        IRequestHandler<SearchCommand, OneOf<ModelWithErrors<SearchCommand>, SearchViewModel>>,
+        IRequestHandler<SelectCommand, OneOf<ModelWithErrors<SearchViewModel>, Success>>
+    {
+        private readonly IAddressSearchService _addressSearchService;
+        private readonly ISearchClient<Onspd> _onspdSearchClient;
+        private readonly JourneyInstance<AddVenueJourneyModel> _journeyInstance;
```
jameslawesynergy, line 47: See previous comments

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+                    .Select(r => new SearchResult()
+                    {
+                        AddressId = r.Id,
+                        StreetAddress = r.StreetAddress
+                    })
+                    .ToArray()
+            };
+        }
+
+        private void ThrowIfFlowStateNotValid()
```
jameslawesynergy, line 170: Unnecessary abstraction

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+                state.Postcode = addressDetail.Postcode;
+                state.Latitude = onspdPostcodeRecord.Record.lat;
+                state.Longitude = onspdPostcodeRecord.Record.@long;
+                state.AddressIsOutsideOfEngland = onspdPostcodeRecord.Record.Country != "England";
+            });
+
+            return new Success();
+        }
+
+        private async Task<SearchViewModel> CreateViewModel(string postcode)
```
jameslawesynergy, line 153: This isn't "creating a view model", it's "performing a post code search". Consider extracting functionality or appropriate naming

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+            var onspdSearchResult = await _onspdSearchClient.Search(new OnspdSearchQuery() { Postcode = request.Postcode });
+
+            if (onspdSearchResult.Items.Count == 0)
+            {
+                return CreateInvalidPostcodeResponse();
+            }
+
+            var vm = await CreateViewModel(request.Postcode);
+
+            if (vm.Results.Count == 0)
```
jameslawesynergy, line 85:
```suggestion
            if (!vm.Results.Any())
```

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+            ThrowIfFlowStateNotValid();
+
+            var validator = new SelectCommandValidator();
+            var validationResult = validator.Validate(request);
+
+            if (!validationResult.IsValid)
+            {
+                // No address specified. Re-run the search with a prompt to select an address.
+                var vm = await CreateViewModel(request.Postcode);
```
jameslawesynergy, line 113: Potential bug - why would you want to re-run the search and recover in this circumstance? Especially without having conducted the onspd postcode verification?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+                return new ModelWithErrors<SearchViewModel>(vm, validationResult);
+            }
+
+            var addressDetail = await _addressSearchService.GetById(request.AddressId);
```
jameslawesynergy, line 117: I figured we wanted to trash this endpoint as it was only required for the legacy journey?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+                    new ValidationFailure(nameof(SelectCommand.AddressId), "Select an address")
+                });
+
+                var vm = await CreateViewModel(request.Postcode);
+                return new ModelWithErrors<SearchViewModel>(vm, badAddressValidationResult);
+            }
+
+            // SearchCommand has already checked that there is an ONSPD entry for this postcode
+            // so there is no valid case where `.Single()` should blow up
+            var onspdSearchResult = await _onspdSearchClient.Search(new OnspdSearchQuery() { Postcode = addressDetail.Postcode });
```
jameslawesynergy, line 135: Why are we running it again? Surely this data could've been added to the journeystate?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
```diff
@@ -0,0 +1,194 @@
...
+            _journeyInstance.ThrowIfCompleted();
+        }
+
+        private class SearchCommandValidator : AbstractValidator<SearchCommand>
+        {
+            public SearchCommandValidator()
+            {
+                RuleFor(c => c.Postcode)
+                    .NotEmpty().WithMessage("Enter a postcode")
+                    .Apply(Rules.Postcode).WithMessage("Enter a real postcode");
```
jameslawesynergy, line 181: "Enter a _real_ postcode"? Surely "Enter a _valid_ postcode" might be more appropriate?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearchResults.cshtml
```diff
@@ -0,0 +1,43 @@
...
+            <p class="govuk-body">
+                <span data-testid="postcode">@Model.Postcode</span>
+                <a asp-action="Index" asp-all-route-data="@RouteValuesHelper.FromQueryString()" asp-fragment="@nameof(Model.Postcode)" class="govuk-link govuk-!-margin-left-5">
+                    Change<span class="govuk-visually-hidden"> postcode</span>
+                </a>
+            </p>
+
+            <govuk-select asp-for="AddressId">
+                <govuk-select-label class="govuk-label--m">Select an address</govuk-select-label>
+                <govuk-select-item value="">@Model.Results.Count address@(Model.Results.Count > 1 ? "es" : "") found</govuk-select-item>
```
jameslawesynergy, line 28: "0 address found" isn't correct English.

```suggestion
                <govuk-select-item value="">@Model.Results.Count address@(Model.Results.Count != 1 ? "es" : "") found</govuk-select-item>
```

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Published.cs
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    public class ViewModel
+    {
+        public string Name { get; set; }
+    }
+
+    public class Handler : IRequestHandler<Query, ViewModel>
+    {
+        private readonly JourneyInstance<AddVenueJourneyModel> _journeyInstance;
```
jameslawesynergy, line 19: See previous comments

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Published.cshtml
```diff
@@ -0,0 +1,35 @@
...
+            <govuk-panel-title>
+                Location added
+            </govuk-panel-title>
+            <span data-testid="VenueName">@Model.Name</span>
+        </govuk-panel>
+
+        <h2 class="govuk-heading-m">What next?</h2>
+        <ul class="govuk-list">
+            <li>
+                <a asp-action="Index" asp-all-route-data="@ProviderContext.RouteValues" class="govuk-link">
```
jameslawesynergy, line 19: I'd rather see an explicit `asp-controller` params in views for safety (no 1-1 mapping between a view and a controller, so no guarantee that the correct action rendered this view)

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/CreateVenue.cs
jameslawesynergy: I think the VM was the wrong choice. We should be aligning to our storage model convention, specifically the SQL storage model convention as that will be the source of truth moving forward. We should also be aiming for consistency across all models (except in the case of the Cosmos storage model as that's impractical to change)

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/CreateVenueHandler.cs
jameslawesynergy: I thought Tim did a PR to update them all the other week? Either way, moving forward we should be using the `UserId`, so using `Email` here is incorrect. And if there's others in other journeys that need changing we either need to get it done or put a ticket in the backlog to do it.

jameslawesynergy: tldr; Correctness trumps consistency.

src/Dfc.CourseDirectory.Web/Controllers/ApprenticeshipsController.cs
jameslawesynergy: Sure, use `string.Empty`.

src/Dfc.CourseDirectory.Web/Views/AddCourse/AddCourseDetails.cshtml
jameslawesynergy: I didn't see a requirement in the story for this - where did it come from? Is this not an opportunity to rip this out rather than creating more tech debt?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Published.cshtml
gunndabad: Yep - that was missed.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
gunndabad: I didn't write the content

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cshtml
gunndabad: Copy is taken from the prototype

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
jameslawesynergy: Ok, why is this optional?

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/CreateVenueHandler.cs
gunndabad: >tldr; Correctness trumps consistency.
I disagree.

Having multiple approaches for a given entity is a mistake, I feel.

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/Queries/CreateVenue.cs
gunndabad: I'll change to `Telephone` for this Create query.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
jameslawesynergy: Surely it would be a ViewModel that inherits from the Command based on previous implementations of this pattern?

src/Dfc.CourseDirectory.Web/Views/AddCourse/AddCourseDetails.cshtml
gunndabad: The notifications were to be unchanged for this story, even if the ACs don't capture it.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: Because non-admins don't have `providerId` query param.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
jameslawesynergy:
```
public Task<IActionResult> Published() =>
    _mediator.SendAndMapResponse<Published.ViewModel, IActionResult>(new Published.Query(), vm => View(vm));
```

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cs
jameslawesynergy: I doubt it's less efficient. And it's certainly more readable.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: That's less readable to my eyes.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cs
jameslawesynergy: I don't think consistency with a different class is an acceptable argument - why does it need to be consistent? In this particular case all it does it make the code more difficult to read.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/Address.cshtml
jameslawesynergy: Missed that, apologies.

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/CreateVenueHandler.cs
jameslawesynergy: I've just looked at the code, including the other Venues journey - everything else is now using `UserId`, not `Email`. Please change this to `UserId`.

src/Dfc.CourseDirectory.Web/Views/AddCourse/AddCourseDetails.cshtml
jameslawesynergy: This seems like a potential gap in refinement to me - is it worth asking the question to PO/UX to get their thoughts?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
jameslawesynergy: So how are we ensuring that non-admins, that don't have a `providerId` querystring, are being subjected to the same validation? that this `requestDataKeys` parameter intends?

In other words, why do we only need this for admin users? Do we even need this functionality at all? Would this functionality be better expressed explicitly in code rather than by virtue of the magic of an action filter?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cshtml
jameslawesynergy: Can we raise this with UX/PO please?

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
jameslawesynergy: I can't find a reference to that copy anywhere on the ticket or in the prototype - could you point me to the right place please?

src/Dfc.CourseDirectory.Core/DataStore/CosmosDb/QueryHandlers/CreateVenueHandler.cs
gunndabad: Doh should've rebased before checking other journeys. Changing...

src/Dfc.CourseDirectory.Web/Views/AddCourse/AddCourseDetails.cshtml
gunndabad: I asked this in the refinement session already.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: That pattern is used where the view model needs more data than is in the `Command`.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cs
gunndabad: Copy/paste from the equivalent in the Edit journey, I suspect. I'll remove it.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
jameslawesynergy: I suppose this is a drawback of the pattern we're using, my only suggestion would be to create a vm wrapper for readability and consistency with the other implementations of this pattern.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: This is required to prevent the provider context from being changed after a journey instance has been started. (If it were changed mid-way through then validation for duplicate venue names could be circumvented, for example.) It's not required for non-admins because they cannot change their provider context (and they will get an error if they try).

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/CheckAndPublish.cs
gunndabad: I find it easier to parse being separate from the `ViewModel` initialization.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: I considered that but I'm not crazy about having the view model properties assigned in multiple places (action vs MediatR handler) and I don't want to thread through the property on the `Query` just to do it in the MediatR handler either.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
gunndabad: In its current form the search results don't return sufficient information to populate an entire address. We could look at amending the types but I'm nervous about extending the scope of this PR any more.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: We don't always have a ViewModel - the Edit journey follows a similar pattern to here where there's just a `Query` and `Command`.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/AddVenueController.cs
gunndabad: This parameter is gone now with the latest changes.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
gunndabad: Amended with latest commit.

src/Dfc.CourseDirectory.WebV2/Features/Venues/AddVenue/PostcodeSearch.cs
gunndabad: Done.

jameslawesynergy (approved):

soharadesign (approved):

PR #1953, opened 2021-03-01
Provider Sql Sync Use AddressItems For Town and County

jameslawesynergy:
Updated Provider SqlSync to use AddressItems for Town and County when values aren't available in Cosmos


jameslawesynergy:
Not sure if we actually want to do this - thoughts?

timabell:
I think this is fine as the data is all sourced from ukrlp and this just smooths over the cracks as we transition from one model to another.

timabell (approved):

PR #1963, opened 2021-03-08
[PTCD-866] Switch VenueHelper to direct-to-cosmos

timabell:
This is one of a few changes needed to bypass the obsolete venue microservice and go direct to cosmos for venue data.

This change is not dependent on the other work in progress or the other changes so is a separate PR.

### DIFF
--- a/src/Dfc.CourseDirectory.Web/Helpers/VenueHelper.cs
+++ b/src/Dfc.CourseDirectory.Web/Helpers/VenueHelper.cs
@@ -1,16 +1,17 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb;
+using Dfc.CourseDirectory.Core.DataStore.CosmosDb.Queries;
 using Dfc.CourseDirectory.Services.Models.Courses;
-using Dfc.CourseDirectory.Services.VenueService;
 
 namespace Dfc.CourseDirectory.Web.Helpers
 {
     public static class VenueHelper
     {
-        public static async Task<Dictionary<Guid, string>> GetVenueNames(IEnumerable<Course> courses, IVenueService venueService )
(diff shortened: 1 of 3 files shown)


gunndabad (approved):

src/Dfc.CourseDirectory.Web/Helpers/VenueHelper.cs
```diff
@@ -19,12 +20,10 @@ public static class VenueHelper
                     {
                         if (!venueNames.ContainsKey((Guid)courseRun.VenueId))
                         {
-                            var result = await venueService.GetVenueByIdAsync(new GetVenueByIdCriteria(courseRun.VenueId.ToString()));
-                            if (result.IsSuccess)
-                            {
-                                Guid.TryParse(result.Value.ID, out Guid venueId);
-                                venueNames.Add(venueId, result.Value.VenueName);
-                            }
+                            var venue = await cosmosDbQueryDispatcher.ExecuteQuery(
```
jameslawesynergy, line 23: This call could potentially be made quite a few times - why not use the `GetVenuesByProvider` query and do the matching in linq, or better yet, create a `GetVenuesByVenueIds` query to grab everything needed in one hit?

jameslawesynergy (approved):

src/Dfc.CourseDirectory.Web/Helpers/VenueHelper.cs
timabell: Do you want to roll that optimisation into this refactor or do a new PR?

jameslawesynergy: It was more of a question of _should we do this?_, rather than a statement of "we _should_ do this" - apologies if that wasn't clear. I've looked at the underlying code and it's doing exactly what you've implemented, albeit via a dumb proxy, so I'm on the fence whether it's worth optimising dying code, especially given that you've technically made it no worse than it already is.

PR #2000, opened 2021-04-13
Ptcd 849 change location to venue

anefiox:
Change location to venue across the site and change some unit tests that were impacted by the text changes.


src/Dfc.CourseDirectory.Core/Models/ApprenticeshipLocationType.cs
```diff
@@ -15,9 +15,9 @@ public static class ApprenticeshipLocationTypeExtensions
         public static string ToDescription(this ApprenticeshipLocationType apprenticeshipLocationType) =>
             apprenticeshipLocationType switch
             {
-                ApprenticeshipLocationType.EmployerBased => "Employer based locations",
-                ApprenticeshipLocationType.ClassroomBased => "Your locations",
-                ApprenticeshipLocationType.ClassroomBasedAndEmployerBased => "Your location and employer locations",
+                ApprenticeshipLocationType.EmployerBased => "Employer based venues",
```
gunndabad, line 18: I don't think these three should be changed. AFAIK we're sticking with 'location' for apprenticeships

src/Dfc.CourseDirectory.Web/Views/Apprenticeships/DeleteDeliveryOption/Index.cshtml
```diff
@@ -4,7 +4,7 @@
 @using Dfc.CourseDirectory.Web.Helpers
 @model Dfc.CourseDirectory.Web.ViewModels.Apprenticeships.DeleteDeliveryOptionViewModel
 @{
-    ViewData["Title"] = "Delete your delivery location"; 
+    ViewData["Title"] = "Delete your delivery venue";
```
gunndabad, line 7: As above, not sure these should be changed

PR #2033, opened 2021-04-29
PTCD-893 -update handlers to stop using cosmos handler to get fechoice …

MrKevJoy:
This is the second PR, that updates the handlers that were previously fetching data from cosmos for fechoices.

- Update anywhere that was fetching fechoice data from cosmos and use the sql provider handler as the two fechoice choice data have been moved onto the providers table.
- update json file to include the two extra properties on providers.json
- updated unit tests.


bucky-uk (approved):

bmatthews-DFE:
Closing due to inactivity

PR #2046, opened 2021-05-12
Overhaul IFileUploadProcessor methods

gunndabad:
Previously, methods on `IFileUploadProcessor` took a `venueUploadId`. Callers would have to run a query to figure out the active `VenueUpload` for a given provider. This is not ideal for a few reasons:
1. `IFileUploadProcessor` manages its own transactions rather than inheriting the current request's transaction. (It has to do this to support operations where multiple processes are working on the upload concurrently like `SaveVenueFile`.) This means it's easy to end up deadlocking (i.e. the query to get the active upload blocks the called method from progressing since they're different transactions). Even a non-deadlocking implementation has a race condition.
2. We end up querying for the `VenueUpload` twice; once to get the active one for the provider and then again inside `IFileUploadProcessor` to check its status etc.. 

This PR changes method signatures to take a `providerId` instead of `venueUploadId`. This simplifies life for callers and removes the deadlock/race.

The exceptions thrown inside `IFileUploadProcessor` have been changed so that they emit an exception that results in a 400 response, should the situation ever occur.


src/Dfc.CourseDirectory.Core/DataStore/Sql/QueryHandlers/GetLatestUnpublishedVenueUploadForProviderHandler.cs
```diff
@@ -1,27 +1,29 @@
...
 using Dapper;
 using Dfc.CourseDirectory.Core.DataStore.Sql.Models;
 using Dfc.CourseDirectory.Core.DataStore.Sql.Queries;
+using Dfc.CourseDirectory.Core.Models;
 
 namespace Dfc.CourseDirectory.Core.DataStore.Sql.QueryHandlers
 {
-    public class GetLatestVenueUploadForProviderWithStatusHandler :
-        ISqlQueryHandler<GetLatestVenueUploadForProviderWithStatus, VenueUpload>
+    public class GetLatestUnpublishedVenueUploadForProviderHandler :
```
gunndabad, line 11: Previous query name wasn't completely accurate; it didn't return the latest venue upload for a given provider and status and the name suggests, it returned the latest venue upload for a given provider *if* it has one of the provided statuses. Changing the name and behaviour here to match.

soharadesign: 👍🏻

MrKevJoy (approved):

PR #2065, opened 2021-06-01
SDD-913 Venue upload inline help

gunndabad:


src/Dfc.CourseDirectory.WebV2/Features/DataManagement/Venues/VenuesDataManagementController.cs
```diff
@@ -206,11 +206,7 @@ public IActionResult DeleteUploadSuccess()
         }
 
         [HttpGet("formatting")]
-        [RequireProviderContext]
```
gunndabad, line 209: Provider context is not required here.

soharadesign: 👍🏻

soharadesign (approved):

PR #2136, opened 2021-07-21
Sdd-931 check lars codes

anefiox:
Many commits in this one due to moving between branches. In hindsight I should have staged changes instead of committing them.


src/Dfc.CourseDirectory.Core/DataManagement/FileUploadProcessor.cs
```diff
@@ -197,11 +199,140 @@ private void CheckStreamIsProcessable(Stream stream)
             }
         }
 
+        //Could put FileMissingLars, FileInvalidLars and FileExpiredLars into one method but they may come in useful elsewhere
```
gunndabad, line 202: I think these checks should be in a single method - `internal async Task<(string[] Invalid, string[] Expired)> CheckLearnAimRefs(Stream stream);`

What's the difference between invalid and missing?

Also the implementation can parse to `CsvCourseRow` instead of dynamic.

src/Dfc.CourseDirectory.Core/DataStore/Sql/Models/Lars.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Dfc.CourseDirectory.Core.DataStore.Sql.Models
+{
+    public class Lars
```
gunndabad, line 5:
```suggestion
    public class LearningDelivery
```

src/Dfc.CourseDirectory.Core/DataStore/Sql/Queries/GetLearningAimRefAndEffectiveTo.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Dfc.CourseDirectory.Core.DataStore.Sql.Queries
+{
+    // Return all LearningAimRefs 
+    public class GetLearningAimRefAndEffectiveTo : ISqlQuery<Lars>
+    {
+        public Lars Lars { get; set; }
+
+        public string LearningAimRef { get; set; }
+    }
```
gunndabad, line 13: This should query all the codes in one go
```suggestion
    public class GetLearningDeliveries : ISqlQuery<IReadOnlyDictionary<string, LearningDelivery>>
    {
        public IEnumerable<string> LearnAimRefs { get; set; }
    }
```

tests/Dfc.CourseDirectory.Core.Tests/DataManagementTests/FileUploadProcessorTests.cs
```diff
@@ -145,5 +146,180 @@ public async Task FileMatchesSchema_RowHasIncorrectColumnCount_ReturnsInvalidRow
             // Assert
             result.Should().Be(FileMatchesSchemaResult.InvalidRows);
         }
+
```
gunndabad, line 149: Each of these tests need to create a file in the valid schema to be valid. Using `DataManagementFileHelper.CreateCourseUploadRows(learnAimRef, rowCount)` may be easier.

tests/Dfc.CourseDirectory.Testing/TestData/TestData.CreateLearningAimRef.cs
```diff
@@ -6,14 +6,15 @@ namespace Dfc.CourseDirectory.Testing
 {
     public partial class TestData
     {
-        public Task<string> CreateLearningAimRef() => WithSqlQueryDispatcher(async dispatcher =>
-        {
-            var learnAimRef = new Random().Next(100000, 109999).ToString("D8");
-
-            await dispatcher.ExecuteQuery(new UpsertLarsLearningDeliveries()
+        public Task<string> CreateLearningAimRef(DateTime? expiredLars = null
```
gunndabad, line 9:
```suggestion
        public Task<string> CreateLearningAimRef(DateTime? effectiveTo = null
```

tests/Dfc.CourseDirectory.WebV2.Tests/FeatureTests/DataManagement/Courses/UploadTests.cs
```diff
@@ -298,5 +298,229 @@ private MultipartFormDataContent CreateMultiPartDataContent(string contentType,
             }
             return content;
         }
+
```
gunndabad, line 301: As above, each of the rows created here need to have all the required columns specified.

src/Dfc.CourseDirectory.Core/DataManagement/FileUploadProcessor.cs
anefiox: I asked @MrKevJoy if I should put them into a single method and he said keep them seperate :-) I happy to change if everyone agrees.

Missing is when a code doesn't exist, Invalid when it's not in the table. I'm going with what errors we have in the prototype.

gunndabad: Missing/invalid makes sense.

Given the implementations are almost identical and they all call a (potentially expensive) SQL call I think it should be one method with one DB call.

anefiox: Do you want me to keep the other methods in or strip them out?

gunndabad: Ideally `FileMissingLars`, `FileInvalidLars`, `FileExpiredLars` are all replaced with `CheckLearnAimRefs`

gunndabad (approved):

PR #2163, opened 2021-07-29
SDD-836 Add missing errors

anefiox:


gunndabad (commented):
I don’t think these messages are correct. They don’t say what’s wrong and don’t align with the others.

src/Dfc.CourseDirectory.Core/Content.resx
```diff
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="ERROR_COURSERUN_ATTENDANCE_PATTERN_NOT_ALLOWED" xml:space="preserve">
-    <value>Attendance pattern</value>
+    <value>Select an attendance pattern</value>
```
gunndabad, line 121: This error is for when attendance pattern is *not* allowed

gunndabad (approved):

PR #2227, opened 2021-08-24
SDD-866 - added resx error messages for apprenticeships

MrKevJoy:
- Added resx error messages for apprenticeships errors

### DIFF
--- a/src/Dfc.CourseDirectory.Core/DataManagement/FileUploadProcessor.Apprenticeships.cs
+++ b/src/Dfc.CourseDirectory.Core/DataManagement/FileUploadProcessor.Apprenticeships.cs
@@ -214,14 +214,17 @@ namespace Dfc.CourseDirectory.Core.DataManagement
                 var rowIsValid = rowValidationResult.IsValid;
                 rowsAreValid &= rowIsValid;
 
+                int.TryParse(parsedRow.StandardCode, out int standardCode);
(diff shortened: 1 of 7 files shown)


src/Dfc.CourseDirectory.Core/DataManagement/FileUploadProcessor.Apprenticeships.cs
```diff
@@ -214,14 +214,17 @@ protected async Task<UploadStatus> GetApprenticeshipUploadStatus(Guid apprentice
                 var rowIsValid = rowValidationResult.IsValid;
                 rowsAreValid &= rowIsValid;
 
+                int.TryParse(parsedRow.StandardCode, out int standardCode);
```
MrKevJoy, line 217: This is required because the story to reject invalid standardcode/standard version has not been done yet. This is here to stop a sql exception trying to convert null to an integer.

gunndabad: Testing shouldn't be using non-valid standard codes until that story is done. I would much rather this were `int.Parse`.

MrKevJoy: changed

gunndabad (approved):

PR #2290, opened 2022-01-31
SDD-1125 changed cancel button to a cancel link

luke-weller:


myselfweller:
Implemented the css change to apply the cancel button styling. Please note the following issues:
- Outdated govuk-frontend meant this required a CSS fix.
- This doesn't have a very good solution for the mobile viewports. It becomes left aligned rather then centered. If we change this to become centered, we run the risk of all links becoming centered.

JoeHS (approved):
Functional testing passed (by Hima in previous weeks) and code review approved - will now be merged into "main" for the next release.

PR #2353, opened 2022-10-14
Merge S24 to "main"

JoeHS:


PurpleDuckMedia (approved):
We don't need the signalr source files. However, they shouldn't cause an issue.

PR #2359, opened 2022-12-02
NCSLT-2259 update signalr libaray version to 7.0.0

luke-weller:


luke-weller:
did not resolve. removing pull request

PR #2392, opened 2023-03-22
Fix for NCSLT-4328 - 

abu-khubaib:
Added update sql query to set Live column to 0 in Pttcd.FindACourseIndex table for a deactivated provider.


abu-khubaib:
This is no more required as the issue has already been fixed by @ZhaomingLu2009 using following pull request
https://github.com/SkillsFundingAgency/dfc-coursedirectory/pull/2386

PR #2428, opened 2023-05-26
Update course when Lars updated

ZhaomingLu2009:


ZhaomingLu2009:
This function app is no longer required as we are fixing the problem by updating the data without invalid characters.

PR #2447, opened 2023-07-24
Bug fix #148873

mimran-DFE:
Bug fix #148873. Work based courses in the same region as the given location should be displayed and should not be impacted by distance.


bmatthews-DFE:
Closing due to inactivity

PR #2463, opened 2023-09-26
Add new tool to generate CSV file from multiple containers of Cosmos DB

abu-khubaib:
Add new tool to generate CSV file from multiple containers of Cosmos DB


gcomptonedu (approved):
Nice one - very handy utility for us to have.

PR #2479, opened 2023-10-16
Remove HTML code from reports course description

zlu-DFE:


bmatthews-DFE:
Closing due to inactivity

PR #2498, opened 2023-12-22
Repoint dfc-devops repository reference to master.

merrilmathew:
Latest commit to dfc-devops to address Az module
import error only available in master branch.


thelious1974:
This has been superceeded by #2499, closing.

PR #2503, opened 2024-01-04
Non LARS Digital Journey

PradeepVoorukonda:


tests/Dfc.CourseDirectory.WebV2.Tests/LayoutTests.cs
```diff
@@ -213,7 +213,7 @@ public async Task AdminUserWithFEAndTLevelsProviderContext_RendersExpectedNav(Te
             }
         }
 
-        [Theory]
+        [Theory(Skip = "Removed courses, venues, tlevels, data upload menu links from provider nav bar")]
```
gcomptonedu, line 216: Should these not be removed if they are no longer needed?

gcomptonedu (changes requested):
Looks good.  Just have two queries.

PR #2582, opened 2024-07-08
Fix for bug 185775

pvoorukonda-DFE:


abu-khubaib (commented):
I had a discussion with Pradeep regarding the script. I recommended the following:

The script should not be included in the post-deploy script but should be executed manually during the release.
Instead of performing a physical delete, the script should implement a soft delete.

PR #2586, opened 2024-08-09
PR for #191735 uri encoding

JoeHS-edu:


Lovre-Schonauer (approved):
New ticket to be created to remove error suppression in this PR after .NET 8 update is completed

PR #2607, opened 2024-11-04
Feature/ad 196037 202848

JHUDSON-SMALL-DFE:


lschonauer-DFE (dismissed):
Approved despite later work this will create. The function apps are re-enabled upon deployment (which rules out disabling them from the Portal) and the 'disable' property was removed in .NET 8 so setting the FAs as HTTP triggers was the only way to ensure automatic triggering stopped.

PR #2658, opened 2025-03-19
Implementation of ArcGIS API

pvoorukonda-DFE:
branch of from main with Implementation of ArcGIS API


src/Dfc.CourseDirectory.Core/ReferenceData/Onspd/OnspdDataImporter.cs
```diff
@@ -30,6 +28,7 @@ public class OnspdDataImporter
         private readonly BlobContainerClient _blobContainerClient;
         private readonly ISqlQueryDispatcher _sqlQueryDispatcher;
         private readonly ILogger<OnspdDataImporter> _logger;
+        private HttpClient _httpClient;
```
hnandra-DFE, line 31: Is it not possible to make the HttpClient readonly here?

src/Dfc.CourseDirectory.Core/ReferenceData/Onspd/OnspdDataImporter.cs
```diff
@@ -39,6 +38,10 @@ public OnspdDataImporter(
             _blobContainerClient = blobServiceClient.GetBlobContainerClient(ContainerName);
             _sqlQueryDispatcher = sqlQueryDispatcher;
             _logger = logger;
+            _httpClient = new HttpClient()
+            {
+                Timeout = TimeSpan.FromSeconds(300)
```
hnandra-DFE, line 43: It is clearer to use TimeSpan.FromMinutes(5) here

hnandra-DFE (changes requested):
Pull request should be pointing to develop

hnandra-DFE:
Do we still require the ImportData() function in this class because it attempts to fetch and import a dataset called "ONSPD_MAY_2023_UK.csv" L23

PR #2659, opened 2025-03-20
Feature/AD-221633

JHUDSON-SMALL-DFE:


mimran-DFE:
Course Directory repository now have 'develop' branch which is currently same as 'main'. 
I would recommend to create new branch from 'develop' instead of 'main'. This is a better approach to avoid accidental merge of changes to main branch.

src/Dfc.CourseDirectory.WebV2/Middleware/NcsSessionCookieMiddleware.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Dfc.CourseDirectory.WebV2.Middleware
+{
+    public class NcsSessionCookieMiddleware
+    {
```
mimran-DFE, line 9: I would recommend to refactor class the code like below:

public class NcsSessionCookieMiddleware
{
    private const string SessionCookieName = "ncs_main_session";
    private const string SessionHeaderName = "X-Ncs-Session-Id";
    private readonly RequestDelegate _requestDelegate;

    public NcsSessionCookieMiddleware(RequestDelegate requestDelegate)
    {
        _requestDelegate = requestDelegate;
    }
    public async Task InvokeAsync(HttpContext context)
    {
        context.Response.OnStarting(() =>
        {
            var sessionId = context.Request.Cookies[SessionCookieName];

            if (string.IsNullOrEmpty(sessionId))
            {
                sessionId = Guid.NewGuid().ToString();
                var options = new CookieOptions
                {
                    Secure = true,
                    IsEssential = true,
                    HttpOnly = true
                };

                context.Response.Cookies.Append(SessionCookieName, sessionId, options);
            }

            context.Response.Headers.Append(SessionHeaderName, sessionId);
            return Task.CompletedTask;
        });
        await _requestDelegate(context);
    }
}

This will make it more readable and maintainable. For example, 

Use string.IsNullOrEmpty. It is safer than checking for null only.
Use Task.CompletedTask. It is preferred over Task.FromResult(0) for void-returning tasks.

mimran-DFE (dismissed):
My recommended changes are implemented successfully. All changes seem to be deployed to DEV successfully and working as expected.

mimran-DFE (approved):

mimran-DFE (approved):

PR #2673, opened 2025-04-11
Added environment appsettings json for lab environment

ajones-DFE:
I have added the environment specific appsettings to the repository which allows the sign in callback to be accepted and stops a 500 error from being raised


abo-olsen-DFE:
While I understand this is just following the existing procedure, I would rather see that we removed the bit in `program.cs` that is making this work, and then add the variables from these files to `appsettings.json` and therefore the `environment variables` on the App Services in Azure. This, I believe, is going to cause a lot less confusion in the future.

PR #2703, opened 2025-06-09
Us 232479 Expired T Level alerts

gcompton-DFE:


src/Dfc.CourseDirectory.Web/Controllers/EditCourse/EditCourseRunController.cs
pvoorukonda-DFE, line 1: Removed sessions to identify non lars courses. is this intentional?

pvoorukonda-DFE (approved):
looks good to me

PR #2739, opened 2025-08-11
Feature/241328 non regulated change text

ibai-gonzalez:


tests/Dfc.CourseDirectory.Core.Tests/ReferenceDataTests/LarsDataImporterTests.cs
```diff
@@ -49,9 +49,9 @@ public async Task ImportData()
             {
                 (await CountSqlRows("LARS.AwardOrgCode")).Should().Be(616);
                 (await CountSqlRows("LARS.Category")).Should().Be(75);
-                (await CountSqlRows("LARS.LearnAimRefType")).Should().Be(127);
-                (await CountSqlRows("LARS.LearningDelivery")).Should().Be(130286);
-                (await CountSqlRows("LARS.LearningDeliveryCategory")).Should().Be(91158);
+                (await CountSqlRows("LARS.LearnAimRefType")).Should().Be(130);
```
ibai-gonzalez, line 52: I fixed this test because it was failing. Still relying on a third party API which is not great to say it mildly :-D

PR #2765, opened 2025-09-26
Applied Changes from Release 94 & 95

pvoorukonda-DFE:
More information about changes introduced in Release 94 & 95 are documented in confluence page - https://skillsfundingagency.atlassian.net/wiki/spaces/DFC/pages/4985749617/PTTCD+-+Investigate+implementation+for+GOV.UK+Frontend+changes+to+align+with+94+95+releases


Dfc.CourseDirectory.WebV2/ModelBinding/DateInputModelConverter.cs
ajones-DFE, line 1: Same here, rather than them importing this redundant code directly, we should look to just use datetime where necessary

Dfc.CourseDirectory.WebV2/ViewModels/TLevels/EditTLevel/EditTLevel.cs
ajones-DFE, line 1: Dateinput is here also

PR #2771, opened 2025-10-03
LARS Automation Changes

mpasumarthi-DFE:


Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
-            }
+                var html = new HtmlDocument(); // using HtmlAgilityPack
+                html.LoadHtml(result);
 
-            async Task DownloadFiles()
-            {
-                _logger.LogInformation("Lars URL- " + _larsDataset.Url);
-                using var resultStream = await _httpClient.GetStreamAsync(_larsDataset.Url);
-                using var zip = new ZipArchive(resultStream);
+                var table = html.DocumentNode.SelectSingleNode("//table[1]");
```
abo-olsen-DFE, line 79: From the code snippet on the ticket

```
// The below always selects the first table, however, instead all tables should be found,
// and the applicable date should be found and considered.
// We want the table where data has the latest applicable date, that is before today.
```

I don't see that this has been considered.

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
+                }
+                else
+                {
+                    downloadDate = "01/01/1900";
+                }
+                var lastDownloadDay = DateOnly.Parse(downloadDate);
+                var baseUrl = _larsDataset.Url;
 
-            await DownloadFiles();
+                var link = "/find-a-learning-aim/DownloadData";
```
abo-olsen-DFE, line 68: Make it configurable. The location may change without the layout.

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
         public async Task ImportData()
         {
-            var extractDirectory = Path.Join(Path.GetTempPath(), "lars");
-            Directory.CreateDirectory(extractDirectory);
+            _logger.LogTrace("LarsDataImport started at: {time}", DateTimeOffset.Now);
+            try
+            {
+                var downloadDate = string.Empty;
+                //Read from blob storage
+                var blobClient = _blobContainerClient.GetBlobClient("larsdownloadinfo.json");
```
abo-olsen-DFE, line 52: Consider making this configurable.

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
+                    var downloadInfoJson = downloadInfoContent.Value.Content.ToString();
+                    var downloadInfo = JsonSerializer.Deserialize<Dictionary<string, string>>(downloadInfoJson);
+                    downloadDate = downloadInfo["LastDownloadDate"];
+                    _logger.LogTrace("Lars last downloaded on {downloadDate} ", downloadDate);
+                }
+                else
+                {
+                    downloadDate = "01/01/1900";
+                }
+                var lastDownloadDay = DateOnly.Parse(downloadDate);
```
abo-olsen-DFE, line 65: Nothing wrong with what is done here, however, I feel that the below is a little more explicit.

```suggestion
                var lastDownloadDay = DateOnly.TryParse(downloadDate, out var res) ? res : new DateOnly(1900, 1, 1);
```

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
-            Task ImportAwardOrgCodeToSql()
-            {
-                var records = ReadCsv<UpsertLarsAwardOrgCodesRecord>("AwardOrgCode.csv");
 
-                return WithSqlQueryDispatcher(dispatcher => dispatcher.ExecuteQuery(new UpsertLarsAwardOrgCodes()
-                {
-                    Records = records
-                }));
-            }
+                    async Task DownloadFiles(string downlaodFile)
```
abo-olsen-DFE, line 127: Nit pick: Misspelling of download.

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
+                        {
+                            Records = records
+                        }));
+                    }
 
-                static bool IsTLevel(UpsertLarsLearnAimRefTypesRecord r) => r.LearnAimRefTypeDesc.StartsWith("T Level", StringComparison.InvariantCultureIgnoreCase);
-            }
+                    async Task<HashSet<string>> ImportCategoryToSql()
+                    {
+                        var records = ReadCsv<UpsertLarsCategoriesRecord>("Category.csv");
```
abo-olsen-DFE, line 158: Configurable?

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,270 @@ public class LarsDataImporter
...
 
-                var includedRecords = records.Where(r => !IsTLevel(r)).ToList();
-                await WithSqlQueryDispatcher(dispatcher =>
-                {
-                    return dispatcher.ExecuteQuery(new UpsertLarsLearningDeliveries()
                     {
-                        Records = includedRecords
-                    });
-                });
+                        const string csv = "LearningDelivery.csv";
```
abo-olsen-DFE, line 186: Configurable?

abo-olsen-DFE (commented):
In the call we had with Lionel, we discussed looking over all tables, and finding the valid one, in case they upload data before it is applicable. I don't see that functionality in the code.

I otherwise think it looks good.
There are quite a few comments around making things configurable, many of which may never change, I will therefore let you decide the best action moving forward, as I also think too many things that can be configured can make it difficult to work with.

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
```diff
@@ -23,185 +28,265 @@ public class LarsDataImporter
...
+                {
+                    var downloadInfoContent = await blobClient.DownloadContentAsync();
+                    var downloadInfoJson = downloadInfoContent.Value.Content.ToString();
+                    var downloadInfo = JsonSerializer.Deserialize<Dictionary<string, string>>(downloadInfoJson);
+                    downloadDate = downloadInfo["LastDownloadDate"];
+                    _logger.LogTrace("Lars last downloaded on {downloadDate} ", downloadDate);
+                }
 
-            await DownloadFiles();
+                var lastDownloadDay = DateOnly.Parse(downloadDate);
```
abo-olsen-DFE, line 63: While I am happy to see that the `else` part is gone. I don't like just using `Parse`. *If* the data in the json file for whatever reason has been changed, then the parsing could fail.

Dfc.CourseDirectory.Core/ReferenceData/Lars/LarsDataImporter.cs
mpasumarthi-DFE: Included the check on Valid from field.

abo-olsen-DFE: Yes, I see that there is a check, but you are not picking the table based on the field. The first table is still the table that is being used. So, in the case that there is a newer version, that is on the first table (the one that we select) but it is not yet applicable, and at the same time, the second table (that is the version we are currently using), has a new upload, that is applicable. We are not getting that new data, but that is what the logic should have found.

